PR #2, opened 2023-02-13
tests: Create symlink sanity to point to a sanity test

spuiuk:
This symlink is to be used to point to a test which is to be used as a sanity test.

This will be used by additional patches to the sit-environment repo to run a sanity test within the CI system.

PR #3, opened 2023-02-14
tests: Add sanity test makefile target

spuiuk:
A different way and a better way to add a sanity test compared to #2. Closing #2 in favour of this approach.


anoopcs9 (approved):
lgtm, thanks.

PR #6, opened 2023-06-29
makefile check fmt rules

synarete:
Developer's makefile rules to check and format Python code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,12 @@
+SELF = $(lastword $(MAKEFILE_LIST))
+ROOT_DIR = $(realpath $(dir $(SELF)))
+
 PYTHONPATH := .
 TEST_INFO_FILE := test-info.yml
 PATH := $(PATH):/usr/local/bin
 export PYTHONPATH TEST_INFO_FILE PATH
+
+MYPY_CMD = $(shell command -v mypy 2> /dev/null)
 
 test:
 	@pytest -v `cat testcases/tests|sed 's/^/testcases\//'`
@@ -10,3 +15,7 @@ sanity_test:
 	@pytest -v testcases/consistency
 
 .PHONY: test sanity_test
+
+.PHONY: check-mypy
+check-mypy:
+	[ ! -z "$(MYPY_CMD)" ] && $(MYPY_CMD) --no-color-output "$(ROOT_DIR)"
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ PATH := $(PATH):/usr/local/bin
 export PYTHONPATH TEST_INFO_FILE PATH
 
 MYPY_CMD = $(shell command -v mypy 2> /dev/null)
+FLAKE8_CMD = $(shell command -v flake8 2> /dev/null)
 
 test:
 	@pytest -v `cat testcases/tests|sed 's/^/testcases\//'`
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
I'm not an active contributor to sit-test-cases at this point, but I have worked with python for quite some time. Take these suggestions with a grain of salt:

* I'm not a fan of pylint. In the past I have found it to be slow, to generate irritating false positives, and I have not enjoyed using it to improve my code. That said, the last time I used was some years ago so it may have improved.
* I do recommend the use of flake8 and black. Do note that the latter means deferring all styling decisions to the tool but I've found that the few style differences I have with the tool I'm willing to let go of for automated consistent styling.
* I would add a rule for checking the formatting/styling rather than just applying it. See the `--check` option.
* I could be mistaken but I think flake8 checks rather than reformats files.
* If this project is going to be primarily a python codebase I'd also recommend the use of tox to run the various tools even if you don't use it for executing the tests. I'd happily discuss with you all how to write a tox.ini (and there are plenty of examples online including sambacc). tox can help handle installing dependencies including the check tools, so you don't need to worry if they are already installed on the system or not.

synarete:
> * flake8

- My recent experience with `pylint` was just fine (with lots of control options via its `pylintrc` config file). Lets give it a try.
- ACK.
- Good idea. I will add `--check`.
- `flake8` is defined as python linter, so yes, I should change the rule to `check-flake8`.
- Fair enough. Should probably be in a separate PR.

Makefile
```diff
@@ -1,5 +1,34 @@
...
+
 test:
 	@ PYTHONPATH=`pwd` ./run_all_tests.sh test-info.yml
 
 sanity_test:
 	@ PYTHONPATH=`pwd` ./run_all_tests.sh test-info.yml sanity_tests
+
+.PHONY: check-pylint
+check-pylint:
+	[ ! -z "$(PYLINT_CMD)" ] && \
```
xhernandez, line 16: If the tool is not installed, shouldn't we fail instead of silently accepting changes that could not comply with the pylint rules ?

The same question can be applied to the other tools.

synarete: @xhernandez On the samba-operator we use the method of installing the tool in locally `.bin` if not found (see https://github.com/samba-in-kubernetes/samba-operator/blob/master/Makefile#L237-L239). We should probably use the same method here, but I did not want to mix too much logic in this PR, so I would implement it in a follow-up PR once we agree on the tool-set we use.

xhernandez: Ok. Thanks.

spuiuk:
From what I understand that the flake8 and pylint are essentially similar tools. I prefer sticking to one for the moment. We can revisit this later once these checkers are run frequently.

synarete:
> From what I understand that the flake8 and pylint are essentially similar tools. I prefer sticking to one for the moment. We can revisit this later once these checkers are run frequently.

Fair enough. Following @phlogistonjohn  comment above, lets go with flake8.

spuiuk (approved):
ACK

spuiuk:
Not immediately committing it immediately to allow others to comment.

Makefile
phlogistonjohn: Please do not copy the approach used in the samba-operator for this repo *if* there's the possibility to use tox. Tox already does this, and it does so in a way that's fairly well documented and widely known (among Python devs).
I would prefer to leave the PR as is or simply remove the conditions, but not reinvent the tox wheel in this case. I am OK with it in samba-operator and samba-containers because samba-operator is a Go(lang) project and Go doesn't have a tox equivalent (that I know of) and both are using a mix of non-language specific tools.

synarete: Fair enough -- I will add a `tox` PR on top of this one. Lets keep this one simple.

PR #7, opened 2023-07-03
smbtorture: add failing samba3.smb2.deny.* tests to flapping tests

spuiuk:
These tests  are being investigated at
https://github.com/gluster/samba-integration/issues/241

and it appears that these are caused by a bug in glusterfs.

### DIFF
--- a/testcases/smbtorture/selftest/flapping.gluster
+++ b/testcases/smbtorture/selftest/flapping.gluster
@@ -7,3 +7,6 @@
 # This is a known fail for samba4 env. We explicitly mark env as
 # samba3 which means we don't ever match with this knownfail.
 ^samba3.smb2.create.quota-fake-file
+
+#https://github.com/gluster/samba-integration/issues/241
+^samba3.smb2.deny.*


spuiuk:
retest this please

spuiuk:
retest this please

spuiuk:
retest this please

spuiuk:
retest this please

anoopcs9:
We may have to use a wildcard for smb2.create too..

spuiuk:
> We may have to use a wildcard for smb2.create too..

The problem is that the failure in the smb2.create.multi is an error setting up the test itself. This is distinct from a failure and cannot be skipped using the flapping test lists.

```
failure: samba3.smb2.deny.deny2 [
vs
error: samba3.smb2.create.multi [
```
From my testing, on my test system, this test is intermittent and seems to be related to some timeout issue.  We need to track this problem in a separate issue.

anoopcs9:
> > We may have to use a wildcard for smb2.create too..
> 
> The problem is that the failure in the smb2.create.multi is an error setting up the test itself. This is distinct from a failure and cannot be skipped using the flapping test lists.
> 
> ```
> failure: samba3.smb2.deny.deny2 [
> vs
> error: samba3.smb2.create.multi [
> ```
> 
> From my testing, on my test system, this test is intermittent and seems to be related to some timeout issue. We need to track this problem in a separate issue.

IIRC, there is an option to specify the timeout(`-t`). Can we try increasing it to 300 instead of default 120?

spuiuk:
removing the patch to flapping.gluster which adds samba3.smb2.create.multi. This cannot be fixed by adding to the flapping list.

spuiuk:
> IIRC, there is an option to specify the timeout(`-t`). Can we try increasing it to 300 instead of default 120?
This wouldn't work. I am investigating what is causing the error and we will probably need to send a patch upstream - but that is for another issue/PR.

anoopcs9 (approved):

PR #9, opened 2023-07-05
Use smbclient for consistency_check

spuiuk:
Use smbclient for consistency checks. By using smbclient, we avoid the need for the test runner to have the necessary permissions to mount the SMB filesystem using a cifs mount. This is important for being able to run the test-cases from a container environment.

I was concerned with speed penalties by moving from cifs mounts to smbclient. However tests show very little difference in speeds.
With cifs mounts: 102.99 seconds
with smbclient: 100.55 seconds.


spuiuk:
The PR requires that the test system has access to smbclient ie. the samba-client package is installed.

spuiuk:
depends on https://github.com/samba-in-kubernetes/sit-environment/pull/11

testcases/consistency/test_consistency.py
```diff
@@ -51,8 +46,8 @@ def consistency_check(mount_point, ipaddr, share_name):
         pytest.fail("Error when open/write/read/close")
 
     finally:
-        if (flag_share_mounted == 1):
-            testhelper.cifs_umount(mount_point)
+        os.unlink(test_file)
+        os.unlink(test_file_resp)
```
xhernandez, line 50: Is it not necessary to also delete `test_file_remote` ?

spuiuk: you are right. The test share otherwise is littered with test files. I have pushed a new patch which calls the rm command right after the get command to fetch the file.

spuiuk:
Setting the Do-not-merge label again to ensure that this code doesn't go in before Shachar's patch with the style fixes goes in.

Depends on: #10

testcases/consistency/test_consistency.py
phlogistonjohn: For a future change, if the files are temporary and don't need to persist across function calls, you might consider switching to a pytest fixture tmp_path: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

spuiuk:
Fresh patches pushed after check with flake8 and black and rebased on the latest tree.

anoopcs9:
retest this please

anoopcs9:
retest this please

testcases/consistency/test_consistency.py
```diff
@@ -26,35 +26,30 @@ def consistency_check(mount_point, ipaddr, share_name):
...
-    except AssertionError:
-        pytest.fail("File content does not match")
-
-    except IOError:
-        pytest.fail("Error when open/write/read/close")
-
+        get_cmds = "get %s %s; rm %s" % (
+            test_file_remote,
+            test_file_resp,
+            test_file_remote,
```
anoopcs9, line 42: Why do we have a 2nd occurrence for _test_file_remote_?

testcases/consistency/test_consistency.py
```diff
@@ -26,35 +26,30 @@ def consistency_check(mount_point, ipaddr, share_name):
...
+        )
+        (ret, output) = testhelper.smbclient(mount_params, get_cmds)
+        assert ret == 0, "Failed to copy file from server"
+        with open(test_file_resp, "r") as f:
+            assert file_content_check(
+                f, test_string
+            ), "File content does not match"
```
anoopcs9, line 49: Can you explain the flow of file creation/copy/deletion here?

testcases/consistency/test_consistency.py
spuiuk: > get test_file_remote test_file_resp
The first one fetches remote:test_file_remote to local:test_file_resp.  

> rm test_file_remote
It then cleans up the remote server in the same smbclient command by deleting remote:test_file_remote.

testcases/consistency/test_consistency.py
spuiuk: The test creates a test_file with a test string. We then copy the file to the server and disconnect from the smb server. Then connect and fetch the file again and compare the contents of the file with out test string. A simple sanity test.

Line 32-33: Create test file with test string
Line 34-36: use smbclient to put the test_file to the smb server.
Line 39: Create the set of commands for the get call - fetch the file and delete the remote file.
44-45:  Run smbclient to fetch the file and then clean up by rm-ing the remote file.
46-49: Compare the file contents on the file just obtained from the remote server and confirm that the contents match.

anoopcs9 (approved):
lgtm, thanks.

testcases/consistency/test_consistency.py
```diff
@@ -26,35 +26,30 @@ def consistency_check(mount_point, ipaddr, share_name):
...
     finally:
-        if flag_share_mounted == 1:
-            testhelper.cifs_umount(mount_point)
+        os.unlink(test_file)
+        os.unlink(test_file_resp)
```
xhernandez, line 52: Just one corner case to consider: if there's an exception while creating the file 'test_file', 'test_file_resp' won't exist and this line will generate another exception.

Probably not critical but I'm not sure if this can confuse pytest about the real error (I don't know enough about pytest).

xhernandez (commented):
Only one comment to consider if it needs to be addressed or not. Othereise, it looks good to me.

testcases/consistency/test_consistency.py
spuiuk: yes. I had thought about it and decided it is not a big deal as it is obviously a problem linked to the earlier issue. With pytest, the two errors will be part of the same output 

Thanks for pointing that out.

testhelper/cmdhelper.py
```diff
@@ -48,3 +49,30 @@ def cifs_umount(mount_point):
     ret = os.system(cmd)
     assert ret == 0, "Error mounting: ret %d cmd: %s\n" % (ret, cmd)
     return ret
+
+
+def smbclient(mount_params, cmd):
+    """Run the following command on smbclient and return the output.
```
obnoxxx, line 55: @spuiuk  to me, it is slightly confusing that this function definition takes a singular "cmd" var as argument while the caller use plural forms they hand in multiple commands as one semicolon-separated string.  It might be good to add a comment explaining this subtlety.

obnoxxx (changes requested):
This looks good in general, but I have one request inline  to add a comment to explain the singular vs plural nature of the cmd argument  to the `smbclient` helper function.

testhelper/cmdhelper.py
spuiuk: Done. 

```
def smbclient(mount_params, cmds):
    """Run the following command on smbclient and return the output.

    Parameters:
    mount_params: Dict containing the mount parameters.
    cmds: String containg the ';' separated commands to run.

    Returns:
    int: Return value from the shell execution
    string: stdout
"""

```

spuiuk:
I have added a new patch which fixes the case when the consistency test is called in a standalone fashion. This is never called in out test setup and hence the bug wasn't caught before. Below is the output from the manual test.
```
# PYTHONPATH=`pwd` ./testcases/consistency/test_consistency.py test-info.yml
Running consistency check:
192.168.123.10 - gluster-vol-replicate
192.168.123.10 - gluster-vol-disperse
192.168.123.11 - gluster-vol-replicate
192.168.123.11 - gluster-vol-disperse
```

testcases/consistency/test_consistency.py
spuiuk: So I decided to add checks for those files before trying to unlink them. This will reduce unnecessary error reports. Please check the latest version of the updates.

obnoxxx (approved):
LGTM, thanks!

PR #10, opened 2023-07-06
python code-style fixups

synarete:
Various code-style fixes to eliminate warning/erro messages from `make check-black` and `make check-flake8`.

### DIFF
--- a/testcases/smbtorture/test_smbtorture.py
+++ b/testcases/smbtorture/test_smbtorture.py
@@ -3,7 +3,8 @@
 # Run smbtorture tests
 
 import testhelper
-import sys, os
+import sys
+import os
 import yaml
 import pytest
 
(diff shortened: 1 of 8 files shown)


spuiuk:
retest this please

testcases/smbtorture/test_smbtorture.py
```diff
@@ -72,7 +77,7 @@ def smbtorture(share_name, test, output):
         filter_subunit_filters,
     )
 
-    format_subunit_cmd = "%s --immediate" % (format_subunit_exec)
+    _ = "%s --immediate" % (format_subunit_exec)
```
phlogistonjohn, line 80: Is this relying on a side effect, it's correct as far as it make the tool happy but I don't see how it accomplishes anything. If we can it may be preferable to simply remove it.

phlogistonjohn (approved):
Approved but with one small nitpick that you can either take or leave.

testcases/smbtorture/test_smbtorture.py
synarete: I did not remove this line completely so we can have this explicit discussion. 
@spuiuk following @phlogistonjohn comment above, is it ok to remove this no-op line?

spuiuk: Not sure why we didn't notice it before. It is a no-op. Please remove it.

I intend on cleaning up the code soon to use the subprocess.run() command.

spuiuk:
retest this please

spuiuk (approved):
Ack 

Tested with the centosci tests and confirmed that the test runs as expected but fails as expected at the usual point of failure in the smbtorture tests.

PR #11, opened 2023-07-06
pyproject: add 'py-modules' entry for 'setuptools'

synarete:
Having multiple packages layout requires explicit 'py-modules' entry for setuptools in order to avoid failures due to "Multiple top-level packages discovered in a flat-layout".

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,9 @@ authors = [
     {name = "Shachar Sharon", email = "ssharon@ibm.com"},
 ]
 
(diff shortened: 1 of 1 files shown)


spuiuk (commented):
ACK - Tested on F38.

spuiuk (approved):

phlogistonjohn (approved):
I didn't try it myself, but it looks ok to me.  We may want to add a note somewhere that this isn't meant to be pip-installable, and should always be run out of the repo.

PR #13, opened 2023-07-18
mypy type hints

synarete:
Allow mypy to make better static type checking by using explicit type hints. 
In order to simplify review, this PR is split into sub-topics but the logic is pretty much the same all over: whenever possible, add explicit type-hints to function arguments and return type to allow mypy (via `make check-mypy` rule) to perform its static code checks. In addition, it also improves code readability.


spuiuk (approved):
ACK.

Looks good.

PR #14, opened 2023-07-20
smbtorture: Use python subprocess module to execute each command

spuiuk:
This is a clean up of the smbtorture test script. 

To run out smbtorture tests, we need to execute smbtorture and pipe it through the filter-subunit and format-subunit commands to return test results. The resulting command string generated is unwieldy.

We clean up the smbtorture tests by using the python subprocess module to pipe the output of each individual command and hopefully make it easier to maintain.

### DIFF
--- a/testcases/smbtorture/test_smbtorture.py
+++ b/testcases/smbtorture/test_smbtorture.py
@@ -8,6 +8,7 @@ import os
 import yaml
 import pytest
 import typing
+import subprocess
 
 script_root = os.path.dirname(os.path.realpath(__file__))
 smbtorture_exec = "/bin/smbtorture"
@@ -16,10 +17,11 @@ format_subunit_exec = script_root + "/selftest/format-subunit"
 smbtorture_tests_file = script_root + "/smbtorture-tests-info.yml"
 
 test_info: typing.Dict[str, typing.Any] = {}
+# Temp filename containing the output of run commands.
 output = testhelper.get_tmp_file("/tmp")
 
 
-def smbtorture(share_name: str, test: str, output: str) -> bool:
+def smbtorture(share_name: str, test: str, tmp_output: str) -> bool:
     # build smbtorture command
     mount_params = testhelper.get_mount_parameters(test_info, share_name)
     smbtorture_cmd = [
@@ -40,7 +42,7 @@ def smbtorture(share_name: str, test: str, output: str) -> bool:
         "/usr/bin/python3",
         filter_subunit_exec,
         "--fail-on-empty",
-        "--prefix='samba3.'",
+        "--prefix=samba3.",
     ]
     for filter in ["knownfail", "knownfail.d"]:
         filter_subunit_cmd.append(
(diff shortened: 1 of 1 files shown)


testcases/smbtorture/test_smbtorture.py
```diff
@@ -45,33 +45,22 @@ def smbtorture(share_name, test, output):
...
-        + "/selftest/flapping.gluster"
-    )
+    subunit_filters = [
+        "knownfail",
+        "knownfail.d",
+        "flapping",
+        "flapping.d",
+        "flapping.gluster",
+    ]
+    filter_subunit_filters = ""
```
phlogistonjohn: While I don't think this is wrong, I generally dislike using strings for something like this. I'd rather use a list like so (some vars abbreviated): `fs = [f"--expected-failures={script_root}/selftest/{filter}" for filter in subs]`.  Then you can join the list later if needed: `fs2 = " ".join(fs)`

testcases/smbtorture/test_smbtorture.py
```diff
@@ -57,19 +59,46 @@ def smbtorture(share_name, test, output):
...
 
-    with open(output, "w") as f:
-        f.write("Command: " + cmd + "\n\n")
+    # run commands - smbtorture
+    smbtorture = subprocess.Popen(
+        smbtorture_cmd, stderr=subprocess.STDOUT, stdout=subprocess.PIPE
+    )
+
+    # run commands - filter_subunit
+    filter_subunit_stdout = open(tmp_output, "w")
```
phlogistonjohn, line 76: Please use open as a context manager (`with open(...) as foo`) as it automatically handles closing the fd, etc.

phlogistonjohn (commented):
Looks OK generally.  Your commit message mentions `subprocess.run` but you are using subprocess.Popen in the code, presumably to use pipes between the processes.

testcases/smbtorture/test_smbtorture.py
phlogistonjohn: Never mind. I see you handle this in subsequent commits.

spuiuk:
The last set of tests had too much output information which included successful tests.
https://jenkins-samba.apps.ocp.cloud.ci.centos.org/job/samba_gluster-integration-test-cases/222/console

The new set of patches addresses this along with the changes recommended by @phlogistonjohn

testcases/smbtorture/test_smbtorture.py
```diff
@@ -97,9 +127,9 @@ def generate_smbtorture_tests(test_info_file):
 )
 def test_smbtorture(share_name, test):
     ret = smbtorture(share_name, test, output)
+    if os.path.exists(output):
+        os.unlink(output)
```
phlogistonjohn, line 131: This is potentially racy (Look before you leap). I think it is preferable to do a:
```suggestion
try:
    os.unlink(output)
except FileNotFoundError:
    pass
````

But this is very minor, if you don't feel like doing this I don't mind.

spuiuk: We are unlikely to hit a race here since only one instance of the runner is expected to call smbtorture. In any case, your suggestion is a good one which can be implemented as a helper function since I've had to implement simple check and unlink for temporary files elsewhere. For now I'll skip this and create a helper to handle deletion of temporary files in another commit.

spuiuk:
retest this please

testcases/smbtorture/test_smbtorture.py
phlogistonjohn: Sure. I also just realized you could also use pathlib here if you want to and the Path.unlink object has a `missing_ok` flag: https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink

spuiuk:
retest this please

spuiuk:
retest this please

anoopcs9 (approved):

PR #22, opened 2023-08-16
mount: run test within subdirectory

spuiuk:
Clean up test artifacts once the test has been run.

### DIFF
--- a/testcases/mount/test_mount.py
+++ b/testcases/mount/test_mount.py
@@ -8,6 +8,7 @@ import os
 import sys
 import pytest
 import typing
+import shutil
 
 from .mount_io import check_io_consistency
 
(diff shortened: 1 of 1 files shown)


testcases/mount/test_mount.py
```diff
@@ -24,8 +25,11 @@ def mount_check(ipaddr: str, share_name: str) -> None:
     try:
         testhelper.cifs_mount(mount_params, mount_point)
         flag_mounted = True
-        check_io_consistency(mount_point)
+        test_dir = mount_point + "/mount_test"
+        os.mkdir(test_dir)
```
synarete, line 29: `test_dit` has type `pathlib.Path`, so better user `test_dir.mkdir()`.

testcases/mount/test_mount.py
```diff
@@ -24,8 +25,11 @@ def mount_check(ipaddr: str, share_name: str) -> None:
...
+        check_io_consistency(test_dir)
     finally:
+        shutil.rmtree(test_dir, ignore_errors=True)
```
synarete, line 32: This line should go under the `if flag_mounted:` condition. In case failure with `testhelper.cifs_mount` we would call this command where `test_dir` is still undefined and get another execption.

testcases/mount/test_mount.py
synarete: That is: 
```
test_dir = pathlib.Path(mount_point) / "mount_test"
test_dir.mkdir()
```

spuiuk: I think we will benefit from a broader application of pathlib in sit-test-cases. We can do this with a 
 larger patch later. I will keep this bit for now.

synarete: Agreed.

anoopcs9 (approved):

PR #25, opened 2023-09-07
mount: Add a stress test with multiple clients

Shwetha-Acharya:
This test simulates multiple clients performing file operations using the concept of threading.

### DIFF
--- a/testcases/mount/mount_stress.py
+++ b/testcases/mount/mount_stress.py
@@ -1,0 +1,57 @@
+import threading
+import testhelper
+import pathlib
+
+
+def _perform_file_operations(
+    client_id: int, root_dir: str, num_operations: int, file_size: int
+) -> None:
+    try:
+        for i in range(num_operations):
+            filename = f"testfile_{client_id}_{i}.txt"
+            file_content = testhelper.generate_random_bytes(file_size)
+            path = pathlib.Path(root_dir, filename)
+            path.write_bytes(file_content)
+            file_content_out = path.read_bytes()
+
+            if file_content_out != file_content:
+                raise IOError("content mismatch")
+
+            path.unlink()
+    except Exception as ex:
+        print(f"Error while stress testing with Client {client_id}: %s", ex)
+        raise
+
+
+def _stress_test(
+    root_dir: str, num_clients: int, num_operations: int, file_size: int
+) -> None:
+    threads = []
+
+    for i in range(num_clients):
+        thread = threading.Thread(
+            target=_perform_file_operations,
+            args=(i, root_dir, num_operations, file_size),
+        )
+        threads.append(thread)
+
+    for thread in threads:
+        thread.start()
+
+    for thread in threads:
+        thread.join()
+
+    print("Stress test complete.")
+
+
+def check_mnt_stress(root_dir: str) -> None:
+    _stress_test(root_dir, num_clients=5, num_operations=20, file_size=2**22)
+    _stress_test(
+        root_dir, num_clients=10, num_operations=30, file_size=2**23
+    )
+    _stress_test(
+        root_dir, num_clients=20, num_operations=40, file_size=2**24
+    )
+    _stress_test(
+        root_dir, num_clients=15, num_operations=25, file_size=2**25
+    )
--- a/testcases/mount/test_mount.py
+++ b/testcases/mount/test_mount.py
@@ -12,6 +12,7 @@ import shutil
 
 from .mount_io import check_io_consistency
 from .mount_dbm import check_dbm_consistency
+from .mount_stress import check_mnt_stress
 
 # Use a global test_info to get a better output when running pytest
 test_info: typing.Dict[str, typing.Any] = {}
@@ -30,6 +31,7 @@ def mount_check(ipaddr: str, share_name: str) -> None:
         os.mkdir(test_dir)
         check_io_consistency(test_dir)
         check_dbm_consistency(test_dir)
+        check_mnt_stress(test_dir)
     finally:
         if flag_mounted:
             shutil.rmtree(test_dir, ignore_errors=True)


testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,46 @@
+import os
+import random
+import string
+import threading
+
+num_clients = 10
+num_operations = 50
+file_size = 1024  # 1 KB
```
synarete, line 8: Two comments:
1) Python supports `2 ** 10` notation (2 to the power of 10), thus no need for extra comment
2) I would vote for later file size in stress-test, say 8M. With such small file-size each thread may end its I/O before the other had a chance to run.

testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _generate_random_string(length: int) -> str:
+    return "".join(random.choice(string.ascii_letters) for _ in range(length))
```
synarete, line 12: This is ok for small files but if enlarging the file size (to 8M) this would be very inefficient. In `mount_io.py` I implemented a pseudo-random bytes sequence generator called `_generate_random_bytes`. Maybe we should make it public in common helper module.

testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,46 @@
...
+            filename = f"testfile_{client_id}_{i}.txt"
+            file_content = _generate_random_string(file_size)
+
+            with open(os.path.join(root_dir, filename), "w") as file:
+                file.write(file_content)
+
+            with open(os.path.join(root_dir, filename), "r") as file:
+                file.read()
+
+            os.remove(os.path.join(root_dir, filename))
```
synarete, line 27: I would prefer using Python's modern pathlib, as it makes more compact code (without duplication). Something like:

```
            path = pathlib.Path(root_dir, filename)
            path.write_text(file_content)
            file_content_out = path.read_text()
            if file_content_out != file_content:
                raise IOError("content mismatch")
            path.unlink()

```


Note that is this code fragment I also check data consistency.

testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,46 @@
...
+
+def stress_test(root_dir: str) -> None:
+    threads = []
+
+    for i in range(num_clients):
+        thread = threading.Thread(
+            target=_perform_file_operations, args=(i, root_dir)
+        )
+        threads.append(thread)
+        thread.start()
```
synarete, line 41: Maybe have another loop for `thread.start` to ensure maximum parallelism. That is, so that no thread will start running while we are still constructing the initial threads-set.

(This one is only nice to have).

testcases/mount/test_mount.py
```diff
@@ -30,6 +31,7 @@ def mount_check(ipaddr: str, share_name: str) -> None:
         os.mkdir(test_dir)
         check_io_consistency(test_dir)
         check_dbm_consistency(test_dir)
+        stress_test(test_dir)
```
synarete, line 34: Consider rename to `check_mt_stress` or something like it (align names with other test functions).

testcases/mount/mount_stress.py
synarete, line 8: Would be nice to have those parameters as arguments. Perhaps even run the test multiple times with different set of args.

testcases/mount/mount_stress.py
Shwetha-Acharya: Sent PR: https://github.com/samba-in-kubernetes/sit-test-cases/pull/27 to address this

Shwetha-Acharya:
I will post the next set of changes corresponding to your comments after the PR: https://github.com/samba-in-kubernetes/sit-test-cases/pull/27 gets merged @synarete

spuiuk:
Thanks for doing this. I am wondering if it is possible to split this out from the current test? The simple mount test which I thought of removing at one time has now evolved to perform io tests and multi mount tests. Maybe we should instead split it into separate io and multimount tests.

Shwetha-Acharya:
> Thanks for doing this. I am wondering if it is possible to split this out from the current test? The simple mount test which I thought of removing at one time has now evolved to perform io tests and multi mount tests. Maybe we should instead split it into separate io and multimount tests.

Yes, that can be done, I can take it up as a next PR

spuiuk:
Having had a quick chat with Shweta, we can consider setting up each of these tests as a separate pytest with a similar setup as described in 
https://stackoverflow.com/questions/7786648/how-to-call-setup-once-for-all-tests-and-teardown-after-all-are-finished

Shwetha-Acharya:
@synarete addressed your comments, though regression will fail (as we have few dependent PRs under review), added next change set to keep this PR open for others to review

Shwetha-Acharya:
retest this please

testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,55 @@
...
+def _perform_file_operations(
+    client_id: int, root_dir: str, num_operations: int, file_size: int
+) -> None:
+    try:
+        for i in range(num_operations):
+            filename = f"testfile_{client_id}_{i}.txt"
+            file_content = testhelper.generate_random_bytes(file_size)
+            path = pathlib.Path(root_dir, filename)
+            path.write_text(file_content.decode())
+            file_content_out = path.read_text()
```
synarete, line 15: Why do you prefer `write_text/read_text` instead of `write_bytes/read_bytes` ? After all, `file_content` is `bytes`, so what is the added value of converting it to text?

testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,55 @@
...
+
+def check_mnt_stress(root_dir: str) -> None:
+    # Case-1: num_clients = 5, num_operations = 20, file_size = 2**22
+    _stress_test(root_dir, 5, 20, 2**22)
+    # Case-2: num_clients = 10, num_operations = 30, file_size = 2**23
+    _stress_test(root_dir, 10, 30, 2**23)
+    # Case-3: num_clients = 15, num_operations = 40, file_size = 2**24
+    _stress_test(root_dir, 10, 40, 2**24)
+    # Case-4: num_clients = 50, num_operations = 50, file_size = 2**25
+    _stress_test(root_dir, 10, 25, 2**25)
```
synarete, line 55: Comments are inconsistent with actual function call. 
Consider using Python's explicit argument names instead of redundant comments. Something like: 
`_stree_test(root_dir, num_clients=5, num_operations=20, file_size=2**22)`

Shwetha-Acharya: I was experimenting with the values, forgot to edit the comments.

testcases/mount/mount_stress.py
Shwetha-Acharya: That makes more sense, updated.

testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,57 @@
...
+    for thread in threads:
+        thread.start()
+
+    for thread in threads:
+        thread.join()
+
+    print("Stress test complete.")
+
+
+def check_mnt_stress(root_dir: str) -> None:
```
phlogistonjohn, line 47: If this were a test we could consider using [pytest's parametrize feature](https://docs.pytest.org/en/latest/parametrize.html) for generating the various combinations.

testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,57 @@
...
+    _stress_test(root_dir, num_clients=5, num_operations=20, file_size=2**22)
+    _stress_test(
+        root_dir, num_clients=10, num_operations=30, file_size=2**23
+    )
+    _stress_test(
+        root_dir, num_clients=20, num_operations=40, file_size=2**24
+    )
+    _stress_test(
+        root_dir, num_clients=15, num_operations=25, file_size=2**25
```
phlogistonjohn, line 56: If I did my math right this means were creating files on the multiple gigabytes range, right?  If so, I'm a little leery of using APIs like `write_bytes` that have to consume a single huge buffer in memory. I would much rather see a loop where the buffer size is something more typical like 4k and then `write()` out each chunk in a loop. In my past experience we had a python program reading files from http into a single huge buffer and that ended up taxing the memory of the system, when changing it to a read-write-loop it was much more manageable.

synarete: @phlogistonjohn I am a bit confused here: the largest file-size here is 32M, and the overall data-set size if up 480M. Those numbers looks reasonably. In all odds, if we want to split those write/reads into multiple chunks, I would vote for larger chunk sized (say, 1M).

phlogistonjohn: I really can't do math. You are right about the sizes.

I'm fine with the code for now, I just don't find the use-one-buffer approach for all file sizes scalable and so I tend to shy away from it. I'd be open to a buffer size of 1M.

testcases/mount/mount_stress.py
Shwetha-Acharya: Added a tracker in [issue-30](https://github.com/samba-in-kubernetes/sit-test-cases/issues/30)

testcases/mount/mount_stress.py
```diff
@@ -0,0 +1,57 @@
...
+def _perform_file_operations(
+    client_id: int, root_dir: str, num_operations: int, file_size: int
+) -> None:
+    try:
+        for i in range(num_operations):
+            filename = f"testfile_{client_id}_{i}.txt"
+            file_content = testhelper.generate_random_bytes(file_size)
+            path = pathlib.Path(root_dir, filename)
+            path.write_bytes(file_content)
+            file_content_out = path.read_text()
```
synarete, line 15: `path.write_bytes` should be paired with `path.read_bytes` (instead of `path.read_text`). We the current code, I have no idea how this test run successfully given that we use random-bytes for `file_content`.

synarete (commented):
Need one final fix.

synarete (approved):
LGTM

Shwetha-Acharya:
@spuiuk @anoopcs9  How do you think about this test with regards to yesterday's discussion? Do we need any more improvements?

anoopcs9:
> @spuiuk @anoopcs9 How do you think about this test with regards to yesterday's discussion about glusterfs issues with multiple clients? Do we need any more improvements?

We have to verify whether multiple kernel CIFS mounts on same client machine can result in multiple _smbd_ processes on the server.

Shwetha-Acharya:
> We have to verify whether multiple kernel CIFS mounts on same client machine can result in multiple _smbd_ processes on the server.
As discussed offline, multiple CIFS mounts will not be in the scope of this test.
Such tests can be added to the performance test suite with bare metal servers

spuiuk:
depends on #29

spuiuk:
We are waiting for PR #29 to be merged to ensure we use the right random string before we merge this particular PR.

spuiuk:
@anoopcs9 , Notice that the test centos-ci/glusterfs says it has passed but in reality the test has failed. Looking at the console output, I notice the following error

```
TASK [test.sit-test-cases : command] *******************************************
Tuesday 17 October 2023  16:13:44 +0000 (0:00:00.037)       0:00:05.650 ******* 
fatal: [client0]: FAILED! => {
    "changed": true,
    "cmd": [
        "make",
        "test"
    ],
    "delta": "1:01:59.717270",
    "end": "2023-10-17 17:15:44.425780",
    "rc": 2,
    "start": "2023-10-17 16:13:44.708510"
}
```

Is this related to the new patches which were merged for sit-environment which captures the logs for the machine?

spuiuk:
I also see the following error

```
_________ test_consistency[192.168.123.10-disperse-glusterfs-default] __________

ipaddr = '192.168.123.10', share_name = 'disperse-glusterfs-default'

    @pytest.mark.parametrize(
        "ipaddr,share_name", generate_consistency_check(test_info_dict)
    )
    def test_consistency(ipaddr: str, share_name: str) -> None:
        tmp_root = testhelper.get_tmp_root()
        mount_point = testhelper.get_tmp_mount_point(tmp_root)
>       consistency_check(mount_point, ipaddr, share_name)

testcases/consistency/test_consistency.py:75: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

mount_point = '/tmp//10924/mnt_0', ipaddr = '192.168.123.10'
share_name = 'disperse-glusterfs-default'

    def consistency_check(mount_point: str, ipaddr: str, share_name: str) -> None:
        mount_params = testhelper.get_mount_parameters(test_info_dict, share_name)
        mount_params["host"] = ipaddr
        try:
            test_file = testhelper.get_tmp_file(mount_point)
            test_file_resp = test_file + ".resp"
            test_file_remote = "test-" + ipaddr + "." + share_name
            with open(test_file, "w") as f:
                f.write(test_string)
            put_cmds = "put  %s %s" % (test_file, test_file_remote)
            (ret, output) = testhelper.smbclient(mount_params, put_cmds)
>           assert ret == 0, "Failed to copy file to server"
E           AssertionError: Failed to copy file to server
E           assert 1 == 0

testcases/consistency/test_consistency.py:36: AssertionError
```

which we should not be seeing. But this doesn't appear to be related to this particular patch. I'll take a closer look tomorrow and will raise separate issues for the two problems identified here.

Shwetha-Acharya:
> @anoopcs9 , Notice that the test centos-ci/glusterfs says it has passed but in reality the test has failed. Looking at the console output, I notice the following error
> 
> ```
> TASK [test.sit-test-cases : command] *******************************************
> Tuesday 17 October 2023  16:13:44 +0000 (0:00:00.037)       0:00:05.650 ******* 
> fatal: [client0]: FAILED! => {
>     "changed": true,
>     "cmd": [
>         "make",
>         "test"
>     ],
>     "delta": "1:01:59.717270",
>     "end": "2023-10-17 17:15:44.425780",
>     "rc": 2,
>     "start": "2023-10-17 16:13:44.708510"
> }
> ```
> 
> Is this related to the new patches which were merged for sit-environment which captures the logs for the machine?

I too have noticed that all the tests are marked with green tick, even though failures are seen in console output

spuiuk:
https://github.com/samba-in-kubernetes/sit-environment/issues/51 reported for CI reporting issues.

spuiuk (approved):
ACK

Issue #26, opened 2023-09-11
 _generate_random_bytes in mount_io.py cannot be reused in other tests


Shwetha-Acharya:
_generate_random_bytes  is an efficient pseudo-random bytes sequence generator. As it is part of mount_io.py it cannot be reused in any other tests.

Making it part of common helper module can solve this issue.

PR #27, opened 2023-09-11
testhelper: make _generate_random_bytes public

Shwetha-Acharya:
_generate_random_bytes is a private function in mount_io.py. Made it public by adding it to testhelper.py. With this change generate_random_bytes can be reused in any number of tests.

Fixes: #26

### DIFF
--- a/testcases/mount/mount_io.py
+++ b/testcases/mount/mount_io.py
@@ -4,9 +4,10 @@
 
 import datetime
 import pathlib
-import random
 import shutil
 import typing
+import testhelper
+import random
 
 
 class DataPath:
@@ -15,10 +16,10 @@ class DataPath:
     def __init__(self, path: pathlib.Path, size: int) -> None:
         self.path = path
         self.size = size
-        self.data = _generate_random_bytes(size)
+        self.data = testhelper.generate_random_bytes(size)
 
     def renew(self) -> None:
-        self.data = _generate_random_bytes(self.size)
+        self.data = testhelper.generate_random_bytes(self.size)
 
     def write(self) -> None:
         self.path.write_bytes(self.data)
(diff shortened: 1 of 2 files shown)


synarete (approved):
LGTM

anoopcs9 (approved):

phlogistonjohn:
I don't care enough to nack this, but I don't really think using this function throughout the code is the best approach. I first recommend surveying what available libraries already exist that might provide a randomized data stream as a pytest fixture.  If no such fixture exists, then I would write one as a fixture.  On top of that I would have the random bytes be available as a standard python type (probably as a readable file-like object or a generator).
Finally, I would prefer to see the random seed explicitly managed - this is the important one and something I think I touched on in the original PR that added this function. It was fine when it was a private func, but it's more important if this starts getting reused all over the place.

synarete:
> I don't care enough to nack this, but I don't really think using this function throughout the code is the best approach. I first recommend surveying what available libraries already exist that might provide a randomized data stream as a pytest fixture. If no such fixture exists, then I would write one as a fixture. On top of that I would have the random bytes be available as a standard python type (probably as a readable file-like object or a generator). Finally, I would prefer to see the random seed explicitly managed - this is the important one and something I think I touched on in the original PR that added this function. It was fine when it was a private func, but it's more important if this starts getting reused all over the place.

@phlogistonjohn Is correct here, and its my fault to fix (probably on-top of this PR). There are 3 different issues here:

0) Have explicit random seed management. Something like [pytest-randomly](https://pypi.org/project/pytest-randomly/)
1) Provide functionality similar to standard `random.randbytes()` where this function does not exist (i.e, CI where python version is less than 3.9)
2) Generate (pseudo) random data for sit-test-cases, without exhaustion the random pool on the host machine or consuming too much CPU (e.g., test 1000 files of 1M random data).

This has been on my TODO list but other things pilled -- I will try to address those soon.

spuiuk (approved):
ACK

spuiuk:
@synarete , if you approve, we can first commit this PR and when we move to CentOS9, we can modify and commit  PR #29  to use library functions to generate random strings.

synarete (approved):
LGTM

PR #28, opened 2023-09-12
pytest with randomly via tox

synarete:
Use tox to run pytest. Use pytest-randomly to have better control on random seed upon re-run due to test failure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /.tox
+__pycache__
--- a/Makefile
+++ b/Makefile
@@ -10,13 +10,13 @@ define runtox
 	@cd "$(ROOT_DIR)" && tox -e $1
 endef
 
+.PHONY: test
 test:
-	@pytest -v `cat testcases/tests|sed 's/^/testcases\//'`
+	$(call runtox, "pytest")
 
+.PHONY: sanity_test
 sanity_test:
-	@pytest -v testcases/consistency
-
-.PHONY: test sanity_test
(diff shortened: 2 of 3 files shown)


phlogistonjohn (approved):
LGTM generally, I'm just not entirely sure why "pytest" and "sanity" are separate envs, since there appears to be overlap between them. I approve, but would like to add a comment/documentation on what each env is meant to accomplish.

synarete:
> LGTM generally, I'm just not entirely sure why "pytest" and "sanity" are separate envs, since there appears to be overlap between them. I approve, but would like to add a comment/documentation on what each env is meant to accomplish.

It preserves the current existing logic in top-level `Makefile` (I did not want to do such cleanup in this PR).

phlogistonjohn:
OK, thanks. I overlooked that was already the case. My general suggestion to document these different cases still stands, but not for this PR. :-)

Shwetha-Acharya (approved):

spuiuk:
retest this please

spuiuk:
@synarete , we may need to install tox on the test system in sit-environment before we can commit this change.

From the console output of the test failure.

```
STDERR:

/bin/sh: tox: command not found
make: *** [Makefile:15: test] Error 127


MSG:

non-zero return code
```

spuiuk (approved):
ACK - but one caveat. We introduce conftest.py but it is never populated. Should we hold on introducing the file until we actually want to add some pytest conf?

PR #29, opened 2023-09-20
mount: use 'random.randbytes' in _generate_random_bytes

synarete:
Now that we may assume Python-version>=3.9, use standard 'random.randbytes' function to generate semi-random I/O testing data. Improve function's doc string; emphasize that this function is defined to avoid random-pool exhaustion on host machine.

### DIFF
--- a/testhelper/testhelper.py
+++ b/testhelper/testhelper.py
@@ -120,20 +120,19 @@ def get_share(test_info: dict, share_num: int) -> str:
 
 def generate_random_bytes(size: int) -> bytes:
     """
-    Creates bytes-sequence filled with random values.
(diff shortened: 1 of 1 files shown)


spuiuk (approved):
ACK

synarete:
> From what I understand, PR #27 will need to be re-based to use this new code. Is this correct?

Yes, but I just realized the CI is still using CentOS8 (with Python 3.6) so we will need also to wait for the transition to CentOS9-stream.

phlogistonjohn (approved):
Looks OK to me.

anoopcs9 (approved):

PR #31, opened 2023-09-27
tox.ini cleanups

synarete:
Few minor cleanups to top-level `tox.ini`. Plus, passenv PYTHONPATH.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,6 +7,7 @@ isolated_build = True
 minversion = 3.9
 passenv =
   TEST_INFO_FILE
+  PYTHONPATH
 deps = pytest
 commands = py.test -v .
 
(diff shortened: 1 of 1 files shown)


phlogistonjohn (commented):
Changes look OK, but I'm a little surprised they are needed. Are they mainly to run better on Windows systems?

synarete:
> Changes look OK, but I'm a little surprised they are needed. Are they mainly to run better on Windows systems?

@spuiuk requested to have `passenv = PYTHONPATH` (reduce tox warning). While at it, I added minor cleanups. 
When executed via windows PowerSehll we should use `tox` directly instead of via wrapper `Makefile` hence it would be nice to ensure that we run at project's root dir.

phlogistonjohn (approved):

spuiuk (approved):
ACK

PR #32, opened 2023-10-02
Allow test with already-mounted smbshare on local dir

synarete:
When running over Windows, we should not execute the cifs_mount/cifs_umount helper functions. Instead, allow running mount tests over already existing smbshare directory, which the user attached via PowerShell's `net use` command.


Shwetha-Acharya (approved):

spuiuk:
commit 31b9e2553511ad79b3a9219111522e18cc312e89
adds support for premounted shares using a different path.

Closing this PR since this support is no longer required.

PR #33, opened 2023-10-12
smbtorture: return None from test_smbtorture

synarete:
Newer pytest expects test entry-point functions to return None. On CI machines (with CentOS9-stream and Python-3.9) it also yields a warn messages.


anoopcs9 (approved):

phlogistonjohn (approved):
lgtm

spuiuk (approved):
ACK

PR #34, opened 2023-10-12
testcases: simplify test structure

Shwetha-Acharya:
pytest by default runs the functions preceeded with test keyword. As a result, all the required test functions will be executed. Hence, this PR simplifies the code by removing the unnecessary main block.

### DIFF
--- a/testcases/consistency/test_consistency.py
+++ b/testcases/consistency/test_consistency.py
@@ -8,7 +8,6 @@
 
 import testhelper
 import os
-import sys
 import pytest
 import typing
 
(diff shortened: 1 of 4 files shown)


phlogistonjohn (approved):
Looks OK to me.

anoopcs9 (changes requested):
lgtm.

I think we have a conflict pending to be resolved.

Shwetha-Acharya:
I just rebased the change after resolving the conflict @anoopcs9

spuiuk (approved):
ACK

Issue #30, opened 2023-09-21
Split multiple test modules in  testcases/mount into individual tests


Shwetha-Acharya:
- [ ] Consider the option of using [setup and teardown](https://stackoverflow.com/questions/7786648/how-to-call-setup-once-for-all-tests-and-teardown-after-all-are-finished) or session fixture.

phlogistonjohn:
+1 to a fixture

Shwetha-Acharya:
Moved parameterized Part of this issue to a new issue

PR #38, opened 2023-10-13
Introduce setup and teardown fixtures

Shwetha-Acharya:
With this change, multiple test modules triggered from testcases/mount can be run as individual tests complying to pytest standards. Also any number of new tests can be added efficiently on the mount.

Old approach:
-  All the different tests are called from test_mount.py:
   https://github.com/samba-in-kubernetes/sit-test-cases/blob/main/testcases/mount/test_mount.py
-  Test_mount.py creates a mount, the required tests are called and then test_mount.py cleans the mount.

New approach :
-  This PR complies with the proposal to use fixtures in order to remove all the load on one single
   module:
   https://github.com/samba-in-kubernetes/sit-test-cases/pull/25#issuecomment-1721320096
   https://github.com/samba-in-kubernetes/sit-test-cases/pull/25#issuecomment-1725320536
-  The tests are no more called from test_mount.py.
-  Setting up of mount and tearing down of mount is taken care by setup and teardown fixture
   which is defined in testcases/mount/conftest.py

   More information about fixtures:
   https://docs.pytest.org/en/6.2.x/fixture.html
-  With this change, test files remain the same, but the initiator function starts with keyword test
   and the each test module starts with keyword 'test'

   That way the tests are called by pytest.
   Usage of setup and teardown fixture becomes easily possible.
   We no more need to to call the test function explictly.

   This also helps with easy debugging and better understanding of the code, as the tests will be
   failed at individual mount/IO test.

   In this case, pytest clearly points which test among test_mount_io.py, test_mount_dbm.py
   and test_mount_stress.py are failing.
   Can be very useful
   - when more and more tests are added
   - when we want to parse test results in sit-environment project

Fixes: #30


Shwetha-Acharya:
This PR need  changes to be merged to pass the tests

anoopcs9:
> This PR need few changes to be merged to pass the tests

Please feel free to edit the description to add the keyword `depends on #<PR.no>` so that its visible in the list of checks.

Shwetha-Acharya:
> > This PR need few changes to be merged to pass the tests
> 
> Please feel free to edit the description to add the keyword `depends on #<PR.no>` so that its visible in the list of checks.

yes sure, I am still working on a change, will update once I push it

anoopcs9:
**centos-ci/xfs** seems to have failed . @Shwetha-Acharya can you check if its related to the proposed change?

conftest.py
```diff
@@ -0,0 +1,48 @@
...
+import pytest
+import os
+import shutil
+import testhelper
+import typing
+
+flag_mounted: bool = False
+tmp_root: str
+test_dir: str
+test_info = os.getenv("TEST_INFO_FILE")
```
spuiuk, line 12: change name to test_info_file for consistency.

conftest.py
```diff
@@ -0,0 +1,48 @@
...
+tmp_root = testhelper.get_tmp_root()
+mount_point = testhelper.get_tmp_mount_point(tmp_root)
+
+
+@pytest.fixture
+def setup_mount(ipaddr: str, sharename: str) -> str:
```
spuiuk, line 18: Do these have to go into conftest.py instead of within test_mount itself?

spuiuk (changes requested):

conftest.py
Shwetha-Acharya: My idea was to define the fixtures inside conftest.py in reference with the common practice.
These fixtures can then be called inside any tests.

spuiuk:
The xfs tests are failing
```
testcases/mount/test_mount_dbm.py::test_check_dbm_consistency[192.168.123.10-share-xfs-default] FAILED [ 96%]
testcases/mount/test_mount_io.py::test_check_io_consistency[192.168.123.10-share-xfs-default] ERROR [100%]

...
E           Exception: Setup failed: Error mounting: ret 256 cmd: mount -t cifs -o username=test1,password=x //[192.168.123.10/share-xfs-default](http://192.168.123.10/share-xfs-default) /tmp//10073/mnt_0

conftest.py:32: Exception
---------------------------- Captured stderr setup -----------------------------
Couldn't chdir to /tmp//10073/mnt_0: No such file or directory

```

I believe that this is happening because of the teardown_mount() call by the first test. That deletes the temporary directory which was created for the test. 

in conftest.py
```
tmp_root = testhelper.get_tmp_root()
mount_point = testhelper.get_tmp_mount_point(tmp_root)
```
will be called only once.

you should either create mount/unmount only once for the entire test or fetch a new tmp_root for every test.

Shwetha-Acharya:
@spuiuk regarding your proposal for removal of some helper functions and that change affecting this PR, I definetly agree there is much more scope for refactoring.

But I propose to keep the PRs independent, and small as possible (mostly addressing one issue)
I propose taking this PR and working on other changes with next PR.

conftest.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
spuiuk, line 1: Can you please move this into the testcases/mount directory for this particular commit. If we do need this in the future, we can move it to a central location.

testcases/consistency/test_consistency.py
```diff
@@ -54,20 +54,8 @@ def consistency_check(mount_point: str, ipaddr: str, share_name: str) -> None:
...
-    arr = []
-    for ipaddr in test_info_file["public_interfaces"]:
-        for share_name in test_info_file["exported_sharenames"]:
-            arr.append((ipaddr, share_name))
-    return arr
-
-
 @pytest.mark.parametrize(
-    "ipaddr,share_name", generate_consistency_check(test_info_dict)
+    "ipaddr,share_name", testhelper.generate_mount_check(test_info_file)
```
spuiuk, line 58: I am not comfortable about moving this pytest parameterize function to a central library. Maybe somewhere which can be included in this test folder itself?

testcases/consistency/test_consistency.py
```diff
@@ -13,8 +13,8 @@
 
 
 test_string = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
-test_info = os.getenv("TEST_INFO_FILE")
-test_info_dict = testhelper.read_yaml(test_info)
+test_info_file = os.getenv("TEST_INFO_FILE")
```
spuiuk, line 16: The convention I have tried to follow in later PRs is
test_info_file -> The filename to be used to read the test_info
test_info -> dict containing the information required to run the test.

Can you please stick to this convention. That will also reduce the large number of changes we see in these commits.

testcases/consistency/test_consistency.py
spuiuk: This will also reduce the scope of these commit to just this test instead of also touching consistency and smbtorture tests. We can consider a helper function later and merge it through a separate PR.

conftest.py
```diff
@@ -0,0 +1,48 @@
...
+        mount_params["host"] = ipaddr
+
+        # mount cifs share
+        testhelper.cifs_mount(mount_params, mount_point)
+        flag_mounted = True
+        test_dir = os.path.join(mount_point, "mount_test")
+        os.makedirs(test_dir, exist_ok=True)
+    except Exception as e:
+        raise Exception(f"Setup failed: {str(e)}")
+    return test_dir
```
spuiuk, line 34: Will this code below the return be executed?

testcases/mount/test_mount_dbm.py
```diff
@@ -89,12 +92,24 @@ def _check_dbm_consistency(base: pathlib.Path, nrecs: int) -> None:
...
+
+
+test_info_file = os.getenv("TEST_INFO_FILE")
+
+
+@pytest.mark.parametrize(
+    "ipaddr,sharename", testhelper.generate_mount_check(test_info_file)
+)
+def test_check_dbm_consistency(
+    ipaddr: str, sharename: str, setup_mount: str
```
spuiuk, line 113: The parameterize function returns two arguments as we did in the older version. 
But we also see a third argument here which is to trigger the mount function. Does this work?

spuiuk (changes requested):

conftest.py
spuiuk: Should you instead use yield in the setup function?
https://docs.pytest.org/en/latest/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization

testcases/mount/test_mount_dbm.py
spuiuk: I have confirmed that this works fine.

testcases/consistency/test_consistency.py
Shwetha-Acharya: yes, I can do it

testcases/consistency/test_consistency.py
Shwetha-Acharya: > I am not comfortable about moving this pytest parameterize function to a central library. Maybe somewhere which can be included in this test folder itself?

Any reason behind it? 
I just thought there were exactly same functions with different names in test as well as in central library. So had thought of reusing the same code.

testcases/mount/test_mount_dbm.py
Shwetha-Acharya: yes, it works as setup_mount is a fixture

testcases/consistency/test_consistency.py
spuiuk: Having given it some more thought. I think I now prefer your approach. Maybe name the call 
testhelper.get_ipaddr_sharename(test_info) 
to match the other calls in testhelper. Similarly for premounted shares, simply use the already existing
testhelper.get_premounted_shares(test_info)

Shwetha-Acharya: I will update that with next rebase

conftest.py
spuiuk: Please ignore this as discussed. This issue has been handled in an other PR.

testcases/consistency/test_consistency.py
spuiuk: Sorry Shwetha, I've had to unresolve this particular change request. 

I was looking at the test output of cephfs test for a different PR
```
testcases/consistency/test_consistency.py::test_consistency[192.168.123.10-share-cephfs-default] PASSED [ 78%]
testcases/consistency/test_consistency.py::test_consistency[192.168.123.11-share-cephfs-default] PASSED [ 81%]
testcases/consistency/test_consistency.py::test_consistency[192.168.123.12-share-cephfs-default] PASSED [ 84%]
testcases/mount/test_mount.py::test_mount[192.168.123.11-share-cephfs-default] PASSED [ 87%]
testcases/mount/test_mount.py::test_mount[192.168.123.10-share-cephfs-default] PASSED [ 90%]
testcases/mount/test_mount.py::test_mount[192.168.123.12-share-cephfs-default] PASSED [ 93%]
```

When I created test_mount, it was the original sanity test which mounted all the public ip addresses to make sure they were working fine. This has been since been done by the sanity tests and the mount test is now more of an io test. 
I wanted to change the test_mount to run the test only on the first ip address instead of all the ip addresses. This means that the parameter generator function will no longer be similar between sanity test(which should still check all ip addresses) and the mount test(which should be renamed).

spuiuk: Can you please leave the test_mount generator function separate and return only the first ip address along with the exported shares.

conftest.py
Shwetha-Acharya: done.

testcases/consistency/test_consistency.py
Shwetha-Acharya: Done!

conftest.py
Shwetha-Acharya: > Will this code below the return be executed?

yes, return is executed. The setup_mount call will be replaced by the created testdir during the execution

Shwetha-Acharya:
I can address next comments, if we are sure we want to go ahead with fixture approach or not

testcases/consistency/test_consistency.py
Shwetha-Acharya: Kept all the generator functions as is

conftest.py
Shwetha-Acharya: > Should you instead use yield in the setup function? https://docs.pytest.org/en/latest/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization

Done!

Shwetha-Acharya:
As per our discussion, went ahead with fixture approach and addressed all the comments and suggestions

Shwetha-Acharya:
@spuiuk Please notice my recent reply on your comment of  "run the test only on the first ip address instead of all the ip addresses." 

I am yet to handle the case of premounted share in this PR, based on your reply to my comment, I can handle that case as well

testcases/mount/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+@pytest.fixture
+def setup_mount(sharename: str) -> typing.Generator[str, None, None]:
+    flag_mounted: bool = False
+    tmp_root = testhelper.get_tmp_root()
+    mount_point = testhelper.get_tmp_mount_point(tmp_root)
+    try:
+        mount_params = testhelper.get_mount_parameters(
+            testhelper.read_yaml(test_info_file), sharename
+        )
+        test_info = testhelper.read_yaml(test_info_file)
```
spuiuk, line 21: Why not set test_info in global variable?

testcases/mount/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+        ipaddr = test_info["public_interfaces"][0]
+        mount_params["host"] = ipaddr
+
+        # mount cifs share
+        testhelper.cifs_mount(mount_params, mount_point)
+        flag_mounted = True
+        test_dir = os.path.join(mount_point, "mount_test")
```
spuiuk, line 28: mount_point as returned by testhelper.get_tmp_mount_point() is of type Path. you could simply do 
test_dir = mount_point / "mount_test"

testcases/mount/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+        os.makedirs(test_dir, exist_ok=True)
```
spuiuk, line 29: test_dir.mkdir()

testcases/mount/conftest.py
spuiuk: you will also need
from pathlib import Path
above

testcases/mount/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+            testhelper.cifs_umount(mount_point)
+        os.rmdir(mount_point)
+        os.rmdir(tmp_root)
+    except Exception as e:
+        raise Exception(f"Teardown failed: {str(e)}")
+
+
+@pytest.fixture
+def generate_mount_check() -> typing.List[typing.Tuple[str, str]]:
+    test_info = testhelper.read_yaml(test_info_file)
```
spuiuk, line 49: global test_info instead of loading the file everytime.

testcases/mount/test_mount_dbm.py
```diff
@@ -89,8 +89,8 @@ def _check_dbm_consistency(base: Path, nrecs: int) -> None:
         db.destroy()
 
 
-def check_dbm_consistency(rootdir: Path) -> None:
-    base = rootdir / "dbm-consistency"
+def test_dbm_consistency(setup_mount):
```
spuiuk, line 92: parameter type.

testcases/mount/test_mount_dbm.py
```diff
@@ -89,8 +89,8 @@ def _check_dbm_consistency(base: Path, nrecs: int) -> None:
...
+    base = Path(setup_mount) / "dbm-consistency"
```
spuiuk, line 93: use Path in parameter type above and you don't need to do this.

testcases/mount/test_mount_io.py
```diff
@@ -99,11 +99,11 @@ def _run_checks(dsets: typing.List[DataPath]) -> None:
         dset.verify_noent()
 
 
-def _check_io_consistency(rootdir: Path) -> None:
+def _check_io_consistency(test_dir: str) -> None:
```
spuiuk, line 102: Use Path instead of str for any path component.

testcases/mount/test_mount_io.py
```diff
@@ -99,11 +99,11 @@ def _run_checks(dsets: typing.List[DataPath]) -> None:
...
     base = None
     try:
         print("\n")
-        base = rootdir / "test_io_consistency"
+        base = Path(test_dir) / "test_io_consistency"
```
spuiuk, line 106: same as above. Stick to Path and you don't have to do this.

testcases/mount/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+    # Yield the setup result
+    yield test_dir
+
+    # Perform teardown after the test has run
+    try:
+        if flag_mounted and test_dir:
+            shutil.rmtree(test_dir, ignore_errors=True)
+            testhelper.cifs_umount(mount_point)
+        os.rmdir(mount_point)
+        os.rmdir(tmp_root)
```
spuiuk, line 42: Can you also use mount_point.rmdir() and tmp_root.rmdir() here. I seem to have missed it in my earlier patch and this is a good time to fix it.

spuiuk (commented):
Some changes requested. 
Also, to complete this patch we need to also handle premounted shares.

testcases/mount/conftest.py
Shwetha-Acharya: Same query https://github.com/samba-in-kubernetes/sit-test-cases/pull/48  applies here as well

testcases/mount/conftest.py
Shwetha-Acharya: Done!

testcases/mount/test_mount_io.py
Shwetha-Acharya: Done!

testcases/mount/test_mount_dbm.py
Shwetha-Acharya: Done!

testcases/mount/test_mount_dbm.py
Shwetha-Acharya: Done!

testcases/mount/conftest.py
Shwetha-Acharya: Done!

testcases/mount/conftest.py
Shwetha-Acharya: Done!

testcases/mount/test_mount_stress.py
```diff
@@ -43,14 +43,16 @@ def _stress_test(
...
-        root_dir, num_clients=20, num_operations=40, file_size=2**24
+        setup_mount, num_clients=10, num_operations=30, file_size=2**23
     )
     _stress_test(
-        root_dir, num_clients=15, num_operations=25, file_size=2**25
+        setup_mount, num_clients=20, num_operations=40, file_size=2**24
+    )
+    _stress_test(
+        setup_mount, num_clients=15, num_operations=25, file_size=2**25
     )
```
spuiuk, line 58: Can you please leave the stress tests the same for now. I intended to ask you to reduce the number of tests here as it takes a long time at the moment. We could reduce the number of tests we run to a single test with 20 clients - I mean if it works with 20 clients, it will probably also work with 5 clients.  We can do this with a later PR.

spuiuk (changes requested):

testcases/mount/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@pytest.fixture
+def setup_mount() -> typing.Generator[Path, None, None]:
+    flag_mounted: bool = False
+    tmp_root = testhelper.get_tmp_root()
+    mount_point = testhelper.get_tmp_mount_point(tmp_root)
+    try:
+        test_info = testhelper.read_yaml(test_info_file)
+        sharename = test_info["exported_sharenames"][0]
```
spuiuk, line 20: I just went through this code again. This will only test one sharename listed. 

The intention is to test with the first listed ip address but test all sharenames which have been provided. The multiple sharenames are needed because we could have different types of volumes in them. eg: Replicated and disperse volumes in gluster. This would only test the first sharename listed ie. replicated. 

This approach is a NAK from me.

spuiuk (changes requested):

testcases/mount/conftest.py
spuiuk: This approach is a NAK

Please consider https://github.com/samba-in-kubernetes/sit-test-cases/pull/38#discussion_r1417905545

testcases/mount/test_mount_stress.py
Shwetha-Acharya: I have not made any changes in stress test as of now, only setup mount is added and hence the indentation is changed by black.

I can add another PR suggested by you.

spuiuk: sorry. yes. that is true

testcases/mount/conftest.py
Shwetha-Acharya: sure, in that case, we might have to import test_info inside sit-test-cases/test-helper/testheper.py from sit-test-cases/mount/conftest.py
from mount.conftest import test_info

Is that what you are suggesting?

testcases/mount/conftest.py
Shwetha-Acharya: Addressed this by making use of only first ipaddr, let me if this works for you.

premounted shares and global test-info are yet to be handled

testcases/mount/conftest.py
spuiuk: No. We don't need to use global variables within the testhelper module.

Please follow the approach used in other tests or even the same test. If adding fixtures require that we use global variables in the testhelper module then I am inclined to NAK the whole patch series.

Shwetha-Acharya: If we are to follow the same approach used in other existing tests, we are already on a right track 

existing approach in main branch: https://github.com/samba-in-kubernetes/sit-test-cases/blob/main/testcases/mount/test_mount.py
current change:https://github.com/samba-in-kubernetes/sit-test-cases/pull/38/commits/70a9c0d39663044fe479960157d77cc9d2d05fed#diff-89e5c02bddad8e8834266d39664b3546c3206ba12c31b972711866e58d51a978

test_info is already a global variable in conftest.py 
Seems we were looking at two different things. I was thinking you did not want me to pass test_info_file to testhelper.read_yaml(test_info_file)

spuiuk: Ahh yes. This is better. There are still a few minor changes I would like you to do and will provide a more detailed review tomorrow.

testcases/consistency/test_consistency.py
spuiuk: several chanes have been made later which address this issue of generator functions for this particular test. Setting this particular change request to resolved.

testcases/mount/conftest.py
spuiuk: The issue of testing all shares has been addressed with the latest version.

testcases/mount/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+import typing
+from pathlib import Path
+
+test_info_file = os.getenv("TEST_INFO_FILE")
+test_info = testhelper.read_yaml(test_info_file)
+
+
+@pytest.fixture
+def setup_mount(
+    first_ipaddr: str, share_name: str
```
spuiuk, line 16: Would you please consider changing the parameter name to ipaddr instead of first_ipaddr. The setup_mount simply mounts the share_name at a ip address and doesn't need to know if this is the first ipaddr. This will only cause confusion for those reading the code later.

This will also require changes below.

testcases/mount/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+) -> typing.Generator[Path, None, None]:
+    flag_mounted: bool = False
+    tmp_root = testhelper.get_tmp_root()
+    mount_point = testhelper.get_tmp_mount_point(tmp_root)
+    try:
+        mount_params = testhelper.get_mount_parameters(
+            testhelper.read_yaml(test_info_file), share_name
```
spuiuk, line 23: just use the global test_info instead of calling testhelper.read_yaml() here.

testcases/mount/test_mount_dbm.py
```diff
@@ -89,8 +91,9 @@ def _check_dbm_consistency(base: Path, nrecs: int) -> None:
         db.destroy()
 
 
-def check_dbm_consistency(rootdir: Path) -> None:
-    base = rootdir / "dbm-consistency"
+@pytest.mark.parametrize("first_ipaddr,share_name", generate_mount_check())
```
spuiuk, line 94: consider renaming first_ipaddr to ipaddr as requested above.

testcases/mount/test_mount_io.py
```diff
@@ -127,6 +129,7 @@ def _reset_random_seed() -> None:
     random.seed(seed)
 
 
-def check_io_consistency(rootdir: Path) -> None:
+@pytest.mark.parametrize("first_ipaddr,share_name", generate_mount_check())
```
spuiuk, line 132: consider renaming first_ipaddr to ipaddr as requested above.

testcases/mount/test_mount_stress.py
```diff
@@ -43,14 +45,17 @@ def _stress_test(
     print("Stress test complete.")
 
 
-def check_mnt_stress(root_dir: Path) -> None:
-    _stress_test(root_dir, num_clients=5, num_operations=20, file_size=2**22)
+@pytest.mark.parametrize("first_ipaddr,share_name", generate_mount_check())
```
spuiuk, line 48: consider renaming first_ipaddr to ipaddr as requested above.

spuiuk (changes requested):
some minor changes. The rest looks good. 

You can also go ahead with re-adding tests for pre-mounted shares after making those changes.

Thanks.

testcases/mount/test_mount_stress.py
Shwetha-Acharya: Done!

testcases/mount/test_mount_dbm.py
Shwetha-Acharya: Done!

testcases/mount/conftest.py
Shwetha-Acharya: Done!

testcases/mount/conftest.py
Shwetha-Acharya: Done!

Shwetha-Acharya:
@spuiuk  as sit environment is not adding any premounted share value inside test-info created by it, I have added logic to ignore premounted share test, if there are no pre mounted shared in test-info fie.

This is useful if any user want to run the tests without the value of premounted_share manually as well

spuiuk (changes requested):

spuiuk:
Please consider the following suggestions  for a future PR. These aren't a result of the changes in this PR but will be nice to have for this test.
1) in _run_dbm_consistency_checks() and _check_io_consistency(), the mkdir call should have parents=False. I don't see why we should have this option set to True. A missing parent directory should be treated as an error.
2) _perform_file_operations() should perform the file operations in a sub directory to allow for easier clean up.

spuiuk:
> @spuiuk sit environment is not adding any premounted share value inside test-info created by it:
> 

> So, inorder to resolve this, I have added logic to ignore premounted share test, if there are no premounted shares in test-info fie.
> 
> This logic to skip is useful even if any user wants to manually run the tests without the value of premounted shares inside the test_info file

This is fine in my opinion. pytest skips the empty tests and it isn't considered an error. I have requested you to remove this logic from the code from my code review which I did earlier.

The premounted shares section is for use with Windows clients where we cannot use mount the shares for the client. In such cases, we need to consider pre-mounted locations to run the tests on. In such cases, we will see the opposite where Skipped messages will be seen on tests which expect to perform the cifs_mounts. eg:  test_check_mnt_stress(), test_dbm_consistency(), test_check_io_consistency()

testcases/mount/test_mount_stress.py
```diff
@@ -43,14 +45,30 @@ def _stress_test(
...
+
+
+@pytest.mark.parametrize("ipaddr,share_name", generate_mount_check())
+def test_check_mnt_stress(setup_mount: Path) -> None:
+    _run_stress_tests(setup_mount)
+
+
+@pytest.mark.parametrize("test_dir", generate_mount_check_premounted())
+def test_dbm_consistency_premounted(
+    test_dir: Path, pre_mounted_shares_available: bool
```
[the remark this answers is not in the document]
Shwetha-Acharya: Done

testcases/mount/test_mount_io.py
```diff
@@ -127,6 +129,20 @@ def _reset_random_seed() -> None:
...
+
+
+@pytest.mark.parametrize("ipaddr,share_name", generate_mount_check())
+def test_check_io_consistency(setup_mount: Path) -> None:
+    _perform_io_consistency_check(setup_mount)
+
+
+@pytest.mark.parametrize("test_dir", generate_mount_check_premounted())
+def test_check_io_consistency_premounted(
+    test_dir: Path, pre_mounted_shares_available: bool
```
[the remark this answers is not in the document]
Shwetha-Acharya: Done

testcases/mount/test_mount_dbm.py
```diff
@@ -89,12 +91,27 @@ def _check_dbm_consistency(base: Path, nrecs: int) -> None:
...
+def test_dbm_consistency(setup_mount: Path) -> None:
+    base = setup_mount / "dbm-consistency"
+    _run_dbm_consistency_checks(base)
+
+
+@pytest.mark.parametrize("test_dir", generate_mount_check_premounted())
+def test_dbm_consistency_premounted(
+    test_dir: Path, pre_mounted_shares_available: bool
+) -> None:
+    if not pre_mounted_shares_available:
```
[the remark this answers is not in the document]
Shwetha-Acharya: Done

testcases/mount/conftest.py
```diff
@@ -0,0 +1,63 @@
...
+            shutil.rmtree(test_dir, ignore_errors=True)
+            testhelper.cifs_umount(mount_point)
+        mount_point.rmdir()
+        tmp_root.rmdir()
+    except Exception as e:
+        raise Exception(f"Teardown failed: {str(e)}")
+
+
+@pytest.fixture
+def pre_mounted_shares_available(test_info):
```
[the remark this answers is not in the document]
Shwetha-Acharya: Done

Shwetha-Acharya:
> Please consider the following suggestions for a future PR. These aren't a result of the changes in this PR but will be nice to have for this test.
> 
> 1. in _run_dbm_consistency_checks() and _check_io_consistency(), the mkdir call should have parents=False. I don't see why we should have this option set to True. A missing parent directory should be treated as an error.
> 2. _perform_file_operations() should perform the file operations in a sub directory to allow for easier clean up.

I will take a look on it after this PR id merged. Else there can be a lot of merge conflicts.

testcases/mount/test_mount_dbm.py
```diff
@@ -89,12 +91,23 @@ def _check_dbm_consistency(base: Path, nrecs: int) -> None:
...
+        _check_dbm_consistency(base_path, 10)
+        _check_dbm_consistency(base_path, 100)
+        _check_dbm_consistency(base_path, 10000)
     finally:
-        shutil.rmtree(base, ignore_errors=True)
+        shutil.rmtree(base_path, ignore_errors=True)
+
+
+@pytest.mark.parametrize("ipaddr,share_name", generate_mount_check())
+def test_dbm_consistency(setup_mount: Path) -> None:
```
spuiuk, line 105: WRT, calling a fixture setup_mount and it receiving the parameters from pytest.mark.parametrize(). 
I can confirm that this is working as expected. I am however unable to find any documentation explaining that this is how it is support to work. If you have a link, can you please point me to their web page which describes this operation.

Thanks

testcases/mount/test_mount_stress.py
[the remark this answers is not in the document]
spuiuk: Sorry, this is unresolved. the parameter here is not required.

spuiuk (changes requested):

testcases/mount/test_mount_stress.py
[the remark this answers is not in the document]
Shwetha-Acharya: Addressed

testcases/mount/test_mount_dbm.py
Shwetha-Acharya: Basically concept of fixtures and parameterize are covered separately in official docs.
Based on their properties, the code is working as expected:

Refer: 

- https://docs.pytest.org/en/7.1.x/how-to/fixtures.html
- https://docs.pytest.org/en/7.3.x/how-to/parametrize.html


Digging deeper I found some instances of such usage:
Refer the example under the title `Parametrization with multiple fixtures` in https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments

spuiuk: So this is a valid way to do it but it isn't documented.
https://github.com/pytest-dev/pytest/discussions/11775

Just for clarity sake, can you please modify the code to use the indirect approach ie. indirect=True
https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization
We should receive a tuple at request.param which can be mapped to ipaddr and share_name and the rest of the code can remain the same.

Shwetha-Acharya: @spuiuk I am not sure if we want to use indirect here.
In our case generate_mount_check is not a fixture. We are simply using the concept of parametrize on a normal function generate_mount_check().

If you see https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization indirect is always used while calling the fixtures inside parametrize. 

I am wondering if are to make  generate_mount_check a fixture inorder to use indirect, which is not looking very right to me.

Please share your views.

spuiuk: setup_mount() is the fixture we wont to run to run setup/teardown for the test.  generate_mount_check() is used to simply generate the list of arguments required for the setup fixture. So we can use it in the same manner as the example given at 
https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization

I believe the code will be similar to the following

The fixture:
```
 @pytest.fixture
 def setup_mount(
-    ipaddr: str, share_name: str
+    request: pytest.FixtureRequest,
 ) -> typing.Generator[Path, None, None]:
+    ipaddr, share_name = request.param
     flag_mounted: bool = False
```

and the tests in a manner similar to
```
-@pytest.mark.parametrize("ipaddr,share_name", generate_mount_check())
+@pytest.mark.parametrize("setup_mount", generate_mount_check(), indirect=True)
 def test_dbm_consistency(setup_mount: Path) -> None:
     base = setup_mount / "dbm-consistency"
```

Shwetha-Acharya: okay, got it now

spuiuk (approved):
ACK

Confirmed that the CI tests for xfs only fail at smb2.create which is a known failure because of selinux issues. All other tests pass as expected.

synarete (approved):
lgtm

PR #39, opened 2023-10-16
Use os.path.join method instead of string concatination

Shwetha-Acharya:
As the seperator('/\') can vary from OS to OS, it is best practice to use os.path.join method instead of string concatination.

This change makes the code platform independent.


phlogistonjohn (approved):
Looks OK to me.

FWIW, You could also consider using pathlib here. Pathlib paths objects support path "concatenation" via the "/" operator, so something like:
```python
mount_point = pathlib.Path(testhelper.get_tmp_mount_point(tmp_root))
test_dir = mount_point / "mount_test"
```

But as a minimal fix this seems fine to me.

anoopcs9 (approved):

PR #41, opened 2023-11-03
Add support for premounted shares

spuiuk:
This is especially useful for windows environments where we cannot use cifs_mount to mount remote shares. At such times, we can premount the shares and use these locations to run our tests.

I have also used this opportunity to convert 
check_io_consistency(test_dir)
check_dbm_consistency(test_dir)
check_mnt_stress(test_dir)

to use Path instead of str as its arguments.


testhelper/testhelper.py
```diff
@@ -136,3 +137,18 @@ def generate_random_bytes(size: int) -> bytes:
...
+
+    Parameters:
+    None
+    Returns:
+    list of paths with shares
+    """
+    arr = []
+    for mounts in test_info["premounted_shares"]:
+        arr.append(Path(mounts))
+    return arr
```
synarete, line 154: More pythonic way of doing it: `return [ Path(mnt) for mnt in test_info["premounted_shares"] ]`

synarete (commented):
I now understand what you had in mind. Looks good (better then my original proposal). Once CI pass, we can take this one.

synarete (approved):
LGTM

Shwetha-Acharya (approved):

PR #42, opened 2023-11-09
Cleanup testhelper and add pathlib.Path to helper functions.

spuiuk:
Several functions are no longer needed after refactoring of some tests.

Also use pathlib.Path so that tests can be used with windows.

depends on #41

PR #44, opened 2023-11-14
Smbtorture output

spuiuk:
The current smbtorture tests report output only in case of test failures. This behaviour was a result of the earlier test runner we used.  With pytest we no longer need to worry about managing the output. 

depends on #42

### DIFF
--- a/testcases/smbtorture/test_smbtorture.py
+++ b/testcases/smbtorture/test_smbtorture.py
@@ -19,9 +19,6 @@ smbtorture_tests_file = script_root + "/smbtorture-tests-info.yml"
 test_info_file = os.getenv("TEST_INFO_FILE")
 test_info = testhelper.read_yaml(test_info_file)
 
-# Temp filename containing the output of run commands.
-output = testhelper.get_tmp_file()
-
 
 def smbtorture(share_name: str, test: str, tmp_output: Path) -> bool:
     # build smbtorture command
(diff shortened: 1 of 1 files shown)


spuiuk:
depends on PR #42

testhelper/cmdhelper.py
```diff
@@ -35,13 +35,20 @@ def cifs_mount(
...
     share = "//" + mount_params["host"] + "/" + mount_params["share"]
-    cmd = "mount -t cifs -o " + mount_options + " " + share + " " + mount_point
+    cmd = (
+        "mount -t cifs -o "
+        + mount_options
+        + " "
+        + share
+        + " "
+        + str(mount_point)
+    )
```
synarete, line 45: Why not simple formatted string? something like: 
```
cmd = f"mount -t cifs -o {mount_options} {share} {mount_pint}"
```

IMHO, it would be more readable.

synarete (approved):
LGTM (but see one minor string-formatting comment).

testhelper/cmdhelper.py
spuiuk: That line was from cmdhelper which was actually committed in an earlier PR. The changes in this PR affect only the test_smbtorture file. I have rebased the series of patches so that only the actual commit remains.

synarete: Fair enough

testcases/smbtorture/test_smbtorture.py
```diff
@@ -19,9 +19,6 @@
 test_info_file = os.getenv("TEST_INFO_FILE")
 test_info = testhelper.read_yaml(test_info_file)
 
-# Temp filename containing the output of run commands.
-output = testhelper.get_tmp_file()
-
```
anoopcs9, line 24: Don't we need it when _smbtorture()_ is called in LINENO:130?

testcases/smbtorture/test_smbtorture.py
```diff
@@ -128,6 +122,7 @@ def generate_smbtorture_tests() -> typing.List[typing.Tuple[str, str]]:
 
 @pytest.mark.parametrize("share_name,test", generate_smbtorture_tests())
 def test_smbtorture(share_name: str, test: str) -> None:
+    output = testhelper.get_tmp_file()
     ret = smbtorture(share_name, test, output)
```
anoopcs9, line 126: In the absence of `output` not defined how does the value look like?

anoopcs9: My bad. It's defined just above .

anoopcs9 (approved):

PR #46, opened 2023-11-20
Makefile: Redirect tox output to a file

anoopcs9:


spuiuk:
Shouldn't this be handled in the sit-environment instead where we store the output into a file which is then copied over. So the output file is both created and consumed by sit-environment.

I would like to see the output shown to the user running sit-test-cases manually. If the redirection has to be done in this repository then I would like to see it pass through tee so that we see the output on both the console and the log file.

spuiuk:
I am curious as to why the centos/xfs test is failing. In this case, because the output is in the log file and the artifact is not shown yet, we can determine what the cause of the failure is. I think we can look at the other PRs for this.

anoopcs9:
> Shouldn't this be handled in the sit-environment instead where we store the output into a file which is then copied over. So the output file is both created and consumed by sit-environment.
> 
> I would like to see the output shown to the user running sit-test-cases manually. If the redirection has to be done in this repository then I would like to see it pass through tee so that we see the output on both the console and the log file.

Fine.

anoopcs9:
Closed in favour of https://github.com/samba-in-kubernetes/sit-environment/pull/67

PR #48, opened 2023-11-30
Optimization of test_info_file and test_info invocation

Shwetha-Acharya:
Currently we are invoking test_info_file and test_info on multiple test modules.

With this change, we are making sure, this invocation takes place only once within testheper.py

Moving all generator functions to testhelper also serves the same advantage.


Shwetha-Acharya:
@spuiuk I have added only two generator functions to test helper as an example.

Just wanted to check with you if you are okay with the idea of moving generator functions to testheper..
If you are okay, I can move rest of the generator functions to testhelper as well

spuiuk:
I would prefer to keep the parameter generator functions along with the test_*() functions which use it.  These are simple functions which I don't mind replicating for each test case as more often than not, these have to be tailored for the individual tests.

Also, I am not a fan of using global variables in helper library calls. While we are using these global variables in each individual test for ease of use reasons, I would like to keep it out of the helper library.

Shwetha-Acharya:
> @spuiuk I have added only two generator functions to test helper as an example.
> 
> Just wanted to check with you if you are okay with the idea of moving generator functions to testheper.. If you are okay, I can move rest of the generator functions to testhelper as well

Ack!
Then we can close this PR.

PR #49, opened 2024-01-09
flapping.cephfs: Add smb2.create.quota-fake-file as knownfail

anoopcs9:
This change replicates what we have with flapping.glusterfs where the mentioned smbtorture subtest is already marked as knownfail in upstream.


spuiuk:
Anoop, We should consider updating the lists from the samba sources directly. Also, I see this test being repeated in flapping.glusterfs and flapping.xfs too. The best place for the test would be in the flapping file but we do not want to use the flapping file since this is copied over from the samba sources. Maybe we want a new flapping file which will be included for all the tests?

spuiuk:
Ack-ed these changes. but we should consider making a separate flapping.common if we have more such tests to add to each of the flapping.<filesystems> list.

Shwetha-Acharya (approved):

anoopcs9:
> test: samba3.smb2.create.quota-fake-file
time: 2024-01-10 11:26:55.890815Z
time: 2024-01-10 11:26:55.901840Z
**xfail**: samba3.smb2.create.quota-fake-file [
Exception: ../../source4/torture/smb2/create.c:3354: status was NT_STATUS_ACCESS_DENIED, expected NT_STATUS_OK: smb2_create failed
]

Above snippet from xfs test [run](https://jenkins-samba.apps.ocp.cloud.ci.centos.org/job/samba_xfs-integration-test-cases/385/artifact/test.out/*view*/) validates the change here.

PR #50, opened 2024-01-10
mount: use pytest.param() on generated parameters

spuiuk:
Using pamertrize with fixtures in pytest means that the test arguments used no longer use the arguments as the test id. Fix this by converting them using pytest.params() with an id argument.

### DIFF
--- a/testcases/mount/conftest.py
+++ b/testcases/mount/conftest.py
@@ -45,12 +45,14 @@ def setup_mount(
         raise Exception(f"Teardown failed: {str(e)}")
 
 
-def generate_mount_check() -> typing.List[typing.Tuple[str, str]]:
+def generate_mount_check() -> typing.List[typing.Any]:
     ipaddr = test_info["public_interfaces"][0]
     exported_sharenames = test_info.get("exported_sharenames", [])
     arr = []
     for share_name in exported_sharenames:
-        arr.append((ipaddr, share_name))
+        arr.append(
+            pytest.param((ipaddr, share_name), id=f"{ipaddr}-{share_name}")
+        )
     return arr
 
 


spuiuk:
@reviewers, I had to use the wildcard typing.Any for the generator function. This is because pytest.ParameterSet is not exported and cannot be used for static type checking. Please let me know if there is a better way to do this.

spuiuk:
To see the difference in reporting.

The mount section of the report 
before this change
```
testcases/mount/test_mount_dbm.py::test_dbm_consistency[setup_mount0] PASSED [  3%]
testcases/mount/test_mount_dbm.py::test_dbm_consistency_premounted[test_dir0] SKIPPED [  6%]
testcases/mount/test_mount_io.py::test_check_io_consistency[setup_mount0] PASSED [ 90%]
testcases/mount/test_mount_io.py::test_check_io_consistency_premounted[test_dir0] SKIPPED [ 93%]
testcases/mount/test_mount_stress.py::test_check_mnt_stress_premounted[test_dir0] SKIPPED [ 96%]
testcases/mount/test_mount_stress.py::test_check_mnt_stress[setup_mount0] PASSED [100%]
```
vs 
after this change
```
testcases/mount/test_mount_dbm.py::test_dbm_consistency_premounted[test_dir0] SKIPPED [  3%]
testcases/mount/test_mount_dbm.py::test_dbm_consistency[192.168.123.10-share-xfs-default] PASSED [  6%]
testcases/mount/test_mount_io.py::test_check_io_consistency_premounted[test_dir0] SKIPPED [  9%]
testcases/mount/test_mount_io.py::test_check_io_consistency[192.168.123.10-share-xfs-default] PASSED [ 12%]
testcases/mount/test_mount_stress.py::test_check_mnt_stress[192.168.123.10-share-xfs-default] PASSED [ 96%]
testcases/mount/test_mount_stress.py::test_check_mnt_stress_premounted[test_dir0] SKIPPED [100%]
```

anoopcs9:
> ```
> testcases/mount/test_mount_dbm.py::test_dbm_consistency_premounted[test_dir0] SKIPPED [  3%]
> testcases/mount/test_mount_dbm.py::test_dbm_consistency[192.168.123.10-share-xfs-default] PASSED [  6%]
> testcases/mount/test_mount_io.py::test_check_io_consistency_premounted[test_dir0] SKIPPED [  9%]
> testcases/mount/test_mount_io.py::test_check_io_consistency[192.168.123.10-share-xfs-default] PASSED [ 12%]
> testcases/mount/test_mount_stress.py::test_check_mnt_stress[192.168.123.10-share-xfs-default] PASSED [ 96%]
> testcases/mount/test_mount_stress.py::test_check_mnt_stress_premounted[test_dir0] SKIPPED [100%]
> ```

Is it like the previous test names like `test_dir0` are insignificant? Isn't there a way to have a combination of both new and old behaviour? Also how different is premounted share test as it still reports in old way?

spuiuk:
> > ```
> > testcases/mount/test_mount_dbm.py::test_dbm_consistency_premounted[test_dir0] SKIPPED [  3%]
> > testcases/mount/test_mount_dbm.py::test_dbm_consistency[192.168.123.10-share-xfs-default] PASSED [  6%]
> > testcases/mount/test_mount_io.py::test_check_io_consistency_premounted[test_dir0] SKIPPED [  9%]
> > testcases/mount/test_mount_io.py::test_check_io_consistency[192.168.123.10-share-xfs-default] PASSED [ 12%]
> > testcases/mount/test_mount_stress.py::test_check_mnt_stress[192.168.123.10-share-xfs-default] PASSED [ 96%]
> > testcases/mount/test_mount_stress.py::test_check_mnt_stress_premounted[test_dir0] SKIPPED [100%]
> > ```
> 
> Is it like the previous test names like `test_dir0` are insignificant? Isn't there a way to have a combination of both new and old behaviour? Also how different is premounted share test as it still reports in old way?

The problem is with the shares which need to be mounted. These are passed through a setup fixture through the request.params pytest object. Hence the parameter names are replaced with setup_mount0, setup_mount1, etc. This particular commit fixes this problem.

The test_dir0 you see is for a skipped test. In this particular test, we pass an empty list since we don't have any premounted shares in this case. You should see the actual directory names when you pass premounted shares.

Shwetha-Acharya:
> Using pamertrize with fixtures in pytest means that the test arguments used no longer use the arguments as the test id. Fix this by converting them using pytest.params() with an id argument.

Could you please elaborate the commit? just explaining how/why test-id was present previously.

spuiuk:
> Could you please elaborate the commit? just explaining how/why test-id was present previously.

The earlier method was to pass parameterized arguments to the test function. These would also serve as test ids which were printed out in the test report allowing readers to exactly understand the test used.

The changes to add setup and teardown function rely on passing the parameters indirectly. Hence the tests id used in the report was by default set to argument_name<n>. This isn't very useful to the reader.

By using the test id with the pytest.param() feature, we now create a testid which is similar to what was reported before the setup/teardown functions were introduced. 

The differences in output can be seen at
https://github.com/samba-in-kubernetes/sit-test-cases/pull/50#issuecomment-1885219980

Shwetha-Acharya (approved):

anoopcs9 (approved):

synarete (approved):

PR #51, opened 2024-01-10
Rename mount test to misc test

spuiuk:
The original tests were simple sanity tests used to test mounting the  SMB share using the cifs mount. These tests have now gone beyond those tests and are now used more as io tests. Renaming the tests to reflect this.


spuiuk:
depends on #51

Shwetha-Acharya (approved):

spuiuk:
Had to change the name from io to misc. This seems to cause problems loading the conftest.py file from the test directory in the form io.conftest which seems to use the inbuilt module instead. By just changing the name to misc, we no longer hit this issue.

synarete (approved):
lgtm

anoopcs9:
Can we have this change rebased on top of latest HEAD?

anoopcs9 (approved):

Issue #54, opened 2024-01-11
Reduce the number of tests in test_mount_stress.py


Shwetha-Acharya:
We could reduce the number of tests we run to a single test with 20 clients - I mean if it works with 20 clients, it will probably also work with 5 clients. 

_Originally posted by @spuiuk in https://github.com/samba-in-kubernetes/sit-test-cases/pull/38#discussion_r1417896362_

PR #55, opened 2024-01-11
test_mount_stress: include single stress test

Shwetha-Acharya:
With this change we are including one stress test with near maximum values i.e num_clients=20, num_operations=40, file_size=2**25

Other existing tests with lower values are eliminated as the goal here should be to test with maximum stress. Tests with lower values are needless when the test with maximum stress passes.

Fixes: #54

### DIFF
--- a/testcases/mount/test_mount_stress.py
+++ b/testcases/mount/test_mount_stress.py
@@ -47,16 +47,7 @@ def _stress_test(
 
 def _run_stress_tests(directory: Path) -> None:
     _stress_test(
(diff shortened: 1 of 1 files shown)


spuiuk (approved):
Ack - Looks good

anoopcs9 (approved):

spuiuk:
Check output of the xfs test to confirm that only 1 failure is reported in smb2.create smbtorture test. This is a known failure.

PR #57, opened 2024-01-17
Update selftests

spuiuk:
The selftest files for smbtorture tests were out of date and needed updating.

We also use this opportunity to exclude the testcases/smbtorture/selftest(and a few more directories which are not part of the repo) from various checks. This means we no longer have to first copy the files from the samba sources and then update them to work with the static checkers. This is far easier to maintain.

PR #58, opened 2024-01-18
flapping.cephfs: Add specific `smb2.session` tests as knownfail

anoopcs9:
Many of the tests from `smb2.session` torture suite are already failing on XFS for which _flapping.xfs_ already contains an entry to skip any such failures. With recent update to selftest entries from upstream, compared to _flapping.xfs_ we are left with only one failure(`smb2.session.reauth4`) which needs to be investigated before adding it to the flapping list.

### DIFF
--- a/testcases/smbtorture/selftest/flapping.cephfs
+++ b/testcases/smbtorture/selftest/flapping.cephfs
@@ -4,3 +4,7 @@
 # This is a known fail for samba4 env. We explicitly mark env as
 # samba3 which means we don't ever match with this knownfail.
 ^samba3.smb2.create.quota-fake-file
+
+# Ignore due to lack of proper multichannel setup.
+^samba3.smb2.session.bind2
+^samba3.smb2.session.two_logoff


spuiuk:
I looked up the old report for the reason to add this to the list.
https://github.com/gluster/samba-integration/issues/118

The problem reported there was specific to glusterfs and we shouldn't see this on xfs an cephfs.. Maybe we open a new issue for this particular problem before we add this to the list. That way we can investigate the issue. 

I also think we should create an issue before we add to any of the flapping.<filesystem> list so that we have some place where we track why this issue exists in the flapping lists over and above what is listed by the samba-team.

anoopcs9:
> I looked up the old report for the reason to add this to the list. [gluster/samba-integration#118](https://github.com/gluster/samba-integration/issues/118)
> 
> The problem reported there was specific to glusterfs and we shouldn't see this on xfs and cephfs.

Interestingly the comment says "All the tests mentioned in the summary pass both for replicated and disperse volumes." but still listed in _testcases/smbtorture/selftest/flapping.glusterfs_.

> Maybe we open a new issue for this particular problem before we add this to the list. That way we can investigate the issue.
>
> I also think we should create an issue before we add to any of the flapping. list so that we have some place where we track why this issue exists in the flapping lists over and above what is listed by the samba-team.

I agree. But in most cases these are added when corresponding tests already fail on XFS.

As far as `smb2.session` test suite is considered I will try to see if some additional configuration is required in first place.

anoopcs9:
> As far as `smb2.session` test suite is considered I will try to see if some additional configuration is required in first place.

Following are the failures seen on XFS:
* `smb2.session.reauth4`: Under investigation.
* `smb2.session.reauth5`: Already present [here](https://github.com/samba-in-kubernetes/sit-test-cases/blob/main/testcases/smbtorture/selftest/knownfail#L214)
* `smb2.session.bind2`: Most likely need to skip because of multichannel requirement
* `samba3.smb2.session.bind_negative_smb3signCtoHd` and `samba3.smb2.session.bind_negative_smb3signHtoCd`: These are already being added as part of updates from #57 
* `samba3.session.two_logoff`: This was added to demonstrate a bug with multichannel setup which further got fixed. We may have to skip the test. See [bug #14532](https://bugzilla.samba.org/show_bug.cgi?id=14532)

spuiuk:
> Interestingly the comment says "All the tests mentioned in the summary pass both for replicated and disperse volumes." but still listed in _testcases/smbtorture/selftest/flapping.glusterfs_.

The tests pass with those entries in flapping.glusterfs. They don't without those entries.

spuiuk (approved):
ACK

anoopcs9:
> > As far as `smb2.session` test suite is considered I will try to see if some additional configuration is required in first place.
> 
> Following are the failures seen on XFS:
> 
>     * `smb2.session.reauth4`: Under investigation.
> 
>     * `smb2.session.reauth5`: Already present [here](https://github.com/samba-in-kubernetes/sit-test-cases/blob/main/testcases/smbtorture/selftest/knownfail#L218)
> 
>     * `smb2.session.bind2`: Most likely need to skip because of multichannel requirement
> 
>     * `samba3.smb2.session.bind_negative_smb3signCtoHd` and `samba3.smb2.session.bind_negative_smb3signHtoCd`: These are already being added as part of updates from [Update selftests #57](https://github.com/samba-in-kubernetes/sit-test-cases/pull/57)
> 
>     * `samba3.session.two_logoff`: This was added to demonstrate a bug with multichannel setup which further got fixed. We may have to skip the test. See [bug #14532](https://bugzilla.samba.org/show_bug.cgi?id=14532)

With #57 merged I've updated the flapping list to be more specific w.r.t individual tests.

spuiuk (approved):
ACK

PR #59, opened 2024-01-18
misc: optimize by avoiding repetitive code

Shwetha-Acharya:
Added a list of testcases to iterate over and to avoid repetition. Also removed unnecessary print statement.

### DIFF
--- a/testcases/misc/test_dbm.py
+++ b/testcases/misc/test_dbm.py
@@ -94,9 +94,10 @@ def _check_dbm_consistency(base: Path, nrecs: int) -> None:
 def _run_dbm_consistency_checks(base_path: Path) -> None:
     base_path.mkdir(parents=True, exist_ok=True)
     try:
-        _check_dbm_consistency(base_path, 10)
-        _check_dbm_consistency(base_path, 100)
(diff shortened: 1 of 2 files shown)


anoopcs9 (approved):

spuiuk:
I am not convinced that these changes are required. This particular code is not expected to change.
The addition of the loop doesn't really add much to functionality or readability in my opinion.

Shwetha-Acharya:
> I am not convinced that these changes are required. This particular code is not expected to change. The addition of the loop doesn't really add much to functionality or readability in my opinion.

I have just made this change according to the basic principle of avoiding repetitive code using loop.
https://github.com/samba-in-kubernetes/sit-test-cases/pull/59/commits/5742b8a88f8fbe432691e834c6631fc43ea97f9c#diff-93cad77f223a70e92e5582538a5f34e8b1ab80b34927d4791694e154f776b941R97  avoids repeatedly calling three functions one after the other.
 

I agree it is not much of functionality, but it is just a clean practice. I am open to hear any opinions.

spuiuk:
I think it negatively impacts the readability of the code.  I vote that we skip this change for now and re-visit it in case these tests get more complicated.

PR #60, opened 2024-01-25
testinfo.yml: reintroduce testinfo.yml.example

spuiuk:
This particular file was incorrectly removed by the patch 
0e09247 (Introduce setup and teardown fixture, 2023-11-22)

Reintroduce this file.


anoopcs9 (approved):
lgtm.

PR #61, opened 2024-01-26
testcases: mark certain tests as privileged

spuiuk:
These tests require privileged access to allow them to mount the shares before we can run the tests on them.

Add a new Makefile target to run only those tests which do not require privileged access.


tox.ini
```diff
@@ -18,6 +18,13 @@ deps =
     pytest-randomly
 commands = pytest -vrfEsxXpP testcases/
 
+[testenv:pytest-nonprivileged]
+deps =
+    pytest
+    pyyaml
+    pytest-randomly
+commands = pytest -vrfEsxXpP -k 'not privileged' testcases/
```
synarete, line 26: Nice -- I was not aware of this pytest option.

synarete (approved):
lgtm

tox.ini
spuiuk: Works well for my containerised dev environment.

tox.ini
```diff
@@ -16,7 +16,14 @@ deps =
     pytest
     pyyaml
     pytest-randomly
-commands = pytest -vrfEsxXpP testcases/consistency testcases/smbtorture testcases/containers testcases/misc
+commands = pytest -vrfEsxXpP testcases/
+
+[testenv:pytest-nonprivileged]
```
anoopcs9, line 21: s/nonpriviliged/unprivileged?

anoopcs9 (changes requested):

tox.ini
spuiuk: So, kubernetes documentation uses unprivileged instead of nonprivileged. So switching to unprivileged.

Makefile
```diff
@@ -30,3 +30,6 @@ PHONY: check-black
 check-black:
 	$(call runtox, "black")
 
+PHONY: tests-unprivileged
+test-unprivileged:
```
anoopcs9, line 34:
```suggestion
.PHONY: test-unprivileged
test-unprivileged:
```

anoopcs9 (changes requested):
Can you please include the following change as an additional commit?
```diff --git a/Makefile b/Makefile
index 1d74be1..a947c2c 100644
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ check-mypy:
 check-flake8:
        $(call runtox, "flake8")
 
-PHONY: check-black
+.PHONY: check-black
 check-black:
        $(call runtox, "black")
```

This slipped through the reviews via #6.

Makefile
spuiuk: fixed.

Makefile
```diff
@@ -30,3 +30,6 @@ PHONY: check-black
 check-black:
 	$(call runtox, "black")
 
+PHONY: tests-non-privileged
```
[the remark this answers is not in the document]
spuiuk: fixed

anoopcs9 (approved):
Thanks.

PR #63, opened 2024-01-31
Add Supplementary group tests

spuiuk:
*** EDIT - Step 1 is now handled in PR #73. Please use this PR only for the supplementary group tests ***
Step 1:
Change the way we describe exports in the test-info.yml file.

1) Introduce exported_shares option. 
```
exported_shares:
  - {"name": "export2"}
  - {"name": "export3", "mntdir": "/mnt-cephfs", "fs": "cephfs"}
```
This allows us to describe both plain exportes smbshares plus premounted shares and the backing filesystem.

2) remove section premounted_shares.
These aren't used in the automated testing yet and can be described through exported_shares section.

Step 2:
Add supplementary group test.

The motivation for making this changes is to cater to a new type of tests which require access to the underlying filesystem outside of SMB/samba. Changes to test-info.yml allow an easy way to describe an exported share where I also give the location of the direct mount so that we can perform these filesystem operations.

The new test added is a Supplementary group test where the following requirements are needed.
1) user defined in test-info.yml should be part of the sg supplementary group.
2) Access to the backend fs outside of SMB is needed - We need to setup a directory to which the supplementary group has write access. We test if the user who is part of the supplementary group has write access.
In this case, we cannot write to the backend using vfs_cephfs module but the write on a filesystem mounted using the ceph-fuse module works fine.

Depends on #73


test-info.yml.example
```diff
@@ -9,12 +9,16 @@ public_interfaces:
 exported_sharenames:
   - "gluster-vol"
 
+# List of dict of exported shares
+# name: share_name
+# mntdir: If this is a premounted share, the mount directory
```
Shwetha-Acharya, line 14: we can rephrase it as mount directory in case of a premounted share

testhelper/testhelper.py
```diff
@@ -76,6 +76,65 @@ def generate_random_bytes(size: int) -> bytes:
...
+    test_info: Dict containing the parsed yaml file.
+    Returns:
+    list of shares
+    """
+    # Instead of recalculating shares at every call,
+    # use saved version in test_info["shares"]
+    if "shares" in test_info:
+        return test_info["shares"]
+
+    default_fs = test_info.get("test_backend", None)
```
Shwetha-Acharya, line 93: Is it necessary to use None here as default value when a key is not present in dict is already None

test-info.yml.example
```diff
@@ -9,12 +9,16 @@ public_interfaces:
 exported_sharenames:
   - "gluster-vol"
 
+# List of dict of exported shares
```
Shwetha-Acharya, line 12: Should this be `exported_shares: list of dict of exported shares`

test-info.yml.example
Shwetha-Acharya, line 14: `mntdir: mount directory in a premounted share` would sound better

testhelper/testhelper.py
```diff
@@ -76,6 +76,65 @@ def generate_random_bytes(size: int) -> bytes:
...
+        if share["name"] == sharename:
+            return share
+    return {}
+
+
+def is_premounted_share(share: dict) -> bool:
+    """
+    Is the share dict a premounted share
+
+    Paramters:
```
Shwetha-Acharya, line 153: typo: parameters

testhelper/testhelper.py
```diff
@@ -76,6 +76,65 @@ def generate_random_bytes(size: int) -> bytes:
...
+    """
+    for share in get_shares(test_info):
+        if share["name"] == sharename:
+            return share
+    return {}
+
+
+def is_premounted_share(share: dict) -> bool:
+    """
+    Is the share dict a premounted share
```
Shwetha-Acharya, line 151: `check if the share is premounted` would sound better

Shwetha-Acharya (commented):
May be you can describe this change in more detail in the commit message for some of us to know better about the need of test for supplementary groups

test-info.yml.example
spuiuk: done

testhelper/testhelper.py
spuiuk: fixed

test-info.yml.example
spuiuk: fixed

test-info.yml.example
spuiuk: fixed

spuiuk:
> May be you can describe this change in more detail in the commit message ([misc: Add test for supplementary groups](https://github.com/samba-in-kubernetes/sit-test-cases/pull/63/commits/08863bc2e5cb7f6cb0c0e5dfb3b1b3f6db29c813) for some of us to know better about the need of test for supplementary groups

Instead of the commit message, I've added it to the python file directly. This allows users to get the information directly from the test files.

testhelper/testhelper.py
spuiuk: fixed

test-info.yml.example
```diff
@@ -9,12 +9,16 @@ public_interfaces:
 exported_sharenames:
   - "gluster-vol"
 
+# exported_shares: List of dict of exported shares
```
Shwetha-Acharya, line 12: Using multi line comment here  similar to testheper.py would be better as it would appear more readable

test-info.yml.example
```diff
@@ -9,12 +9,16 @@ public_interfaces:
...
+# name: share_name
```
Shwetha-Acharya, line 13: just curious why we would need a separate exported_sharenames and again a name exported_shares.  

Let me know if they are any different or there is any special use case? If not could we just use share name from exported_shares?

Shwetha-Acharya (commented):
Added few comments.
Rest of the implementation looks good to me.

test-info.yml.example
spuiuk: They have 2 different formats in the way they are expressed in the yaml. exported_sharenames is how it is currently reported but it only lists the shares exported from that sharename which is inadequate for a complex test. This is currently used by sit-environment to report the shares but in the future we want to move to providing additional information in exported_shares which the current exported_sharename field doesn't provide.

Since we have multiple projects, we have a chicken and egg problem. We can't make changes to one which will break the  testing on the other. So for now, the changes ensure that both exported_sharenames and exported_shares are supported. Once these changes go in, I will put in a PR to move sit-environment to use exported_shares instead of exported_sharenames. Once changes go into both, we can consider removing support for exported_sharenames.

test-info.yml.example
spuiuk: This is an example file for yaml content. Yaml doesn't allow multi line comments like in Python.

xhernandez (commented):
Since SIT environment is becoming more configurable now, and it could provide more complex configurations in the near future, I think we should also make the test-info.yml configuration more flexible.

I would nest all configuration below each share, so that different share can have different configurations.

For example:

```yaml
shares:
  share-cephfs-vfs:
    # If present, it means share is pre-mounted
    path: /mnt/share/share-cephfs-vfs
    backend:
      name: cephfs
      variant: vfs
      # If present, this provides direct access to the backend, bypassing Samba
      path: /mnt/backend/share
    users:
      test1: x
      test2: x
    addresses:
      - 192.168.123.10
      - 192.168.123.11
```

spuiuk:
Thanks, As discussed on chat, we will also add options to add default options which will make it easier to write the test-info.yml files manually.

spuiuk:
I have uploaded the latest changes. The supplementary tests will not be automatically run in the sit-environment since we have not setup backend filesystems and the sit-environment will have to be modified to use the new format for test-info.yml before we can run the supplementary tests.

However the changes introduced here changes the way we layout the environment in test-info.yml.

spuiuk:
depends on #73

spuiuk:
Split this PR with the changes to test-info.yml now handled in #73. Please use this PR to comment on supplementary tests.

PR #70, opened 2024-03-12
smbtorture/selftest: Add flapping.gpfs

anoopcs9:
Use a dedicated flapping list similar to what we have for XFS to start with filtering GPFS smbtorture test run results.


Shwetha-Acharya (approved):

spuiuk (approved):
ACK

PR #73, opened 2024-03-23
test-info.yml changes

spuiuk:
Splitting this set of changes out of PR #63.

These commits change the way we describe shares in test-info.yml file.

These changes will allow for us to set server, backend fs type and users for each individual share. This means that shares across multiple smb shares and filesystems can be tested at the same time.

```
# Server hostname to be tested
server: server_name

# Users to use for authentication
users:
  user1: user1password

# Backend filesystem of the exported shares
backend: glusterfs

# shares: List of dict of exported shares
shares:
  # share name export1
  export1:
    # If present, it means the share is pre-mounted
    path: /mnt/share/export1-cephfs-vfs
    backend:
      # If present backend filesystem
      name: cephfs.vfs
    # If present, username to perform the tests with on this share
    users:
      test2: x
    # If present, the hostname to use to connect to the test server
    server: hostname1
  # share name export2
  export2:
    # Use default values set for this share
```

With these changes, we can still continue using the old format to describe test-info.yml but they will be deprecated and removed once we have made the changes to the sit-environment to use the new format.


testhelper/testhelper.py
```diff
@@ -4,17 +4,59 @@
...
+        else:
+            shares[share]["name"] = share
+        if "backend" not in shares[share]:
+            shares[share]["backend"] = {}
+        if "name" not in shares[share]["backend"]:
+            shares[share]["backend"]["name"] = default_backend
+        if "server" not in share:
+            shares[share]["server"] = default_server
+        if "users" not in share:
+            shares[share]["users"] = default_users
```
synarete, line 56: Perhaps better use Python's [dict.setdefault](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) ? Something like (untested): `shares[share].setdefault("backend", {})`

synarete, line 56: This logic deserve its own (private) function.

testhelper/testhelper.py
```diff
@@ -4,17 +4,59 @@
...
+        or test_info.get("public_interfaces", ["localhost"])[0]
+    )
+    default_users = test_info.get("users")
+    if default_users is None:
+        users = test_info.get("test_users", None)
+        if users is None:
+            default_users = {}
+        else:
+            for user in users:
+                default_users = {user["username"]: user["password"]}
```
synarete, line 42: This logic deserves its own function.

testhelper/testhelper.py
```diff
@@ -85,8 +170,11 @@ def get_premounted_shares(test_info: dict) -> typing.List[Path]:
     Returns:
     list of paths with shares
     """
-    premounted_shares = test_info.get("premounted_shares", [])
-    return [Path(mnt) for mnt in premounted_shares]
+    arr = []
+    for share in get_shares(test_info).values():
+        if is_premounted_share(share):
+            arr.append(Path(share["path"]))
+    return arr
```
synarete, line 177: Python style:
```
share_values = get_shares(test_info).values()
return [Path(s["path") for s in share_values if is_premounted_share(s)]
```

synarete (commented):
The overall logic is fine, but needs a bit of (Python style) cleanups.

testhelper/testhelper.py
spuiuk: done. This cleans up the code nicely.

spuiuk: I think the cleanup after using the dict.setdefault() means we don't really need a new function for this which would have been only used once. So I've cleaned up this code but not split it out into a new function. I hope that is ok.

testhelper/testhelper.py
spuiuk: done

testhelper/testhelper.py
spuiuk: done

selftest/test_testhelper.py
```diff
@@ -0,0 +1,62 @@
...
+    premounted = testhelper.get_premounted_shares(testinfo)
+    assert len(premounted) == 1
+    assert Path("/mnt/share/export1-cephfs-vfs") in premounted
+
+
+def test_generate_exported_shares():
+    testinfo = testhelper.read_yaml("test-info1.yml")
+    arr = []
+    for share in testhelper.get_exported_shares(testinfo):
+        s = testhelper.get_share(testinfo, share)
```
Shwetha-Acharya, line 59: I am not very comfortable with one letter variable name, I would vote for something more descriptive like share_info

testhelper/testhelper.py
```diff
@@ -4,17 +4,66 @@
...
+
+def _get_default_users(test_info: dict) -> typing.Dict:
+    default_users = test_info.get("users")
+    if default_users is None:
+        default_users = {}
+        users = test_info.get("test_users")
+        if users is not None:
+            for user in users:
+                default_users[user["username"]] = user["password"]
+    return default_users
```
synarete, line 26: Consider this alternative form:
```
default_users = test_info.get("users", {})
if len(default_users) == 0):
    for user in test_info.get("test_users", []):
        default_users[user["username"]] = user["password"]
return default_users
```

(or something like it).

testhelper/testhelper.py
```diff
@@ -4,17 +4,66 @@
...
+
+
+def _get_default_server(test_info: dict) -> str:
+    return (
+        test_info.get("server")
+        or test_info.get("public_interfaces", ["localhost"])[0]
+    )
+
+
+def _get_default_users(test_info: dict) -> typing.Dict:
```
synarete, line 18: I think you can have more fine-grained type annotation here (`typing.Dict[str, str]`).

synarete (commented):
Looks fine, but few style issues.

testhelper/testhelper.py
spuiuk: done

testhelper/testhelper.py
spuiuk: done

synarete (approved):
lgtm

selftest/test_testhelper.py
Shwetha-Acharya: Thanks for the update, I still see some occurrences of 's' in testhelper/testhelper.py
Please see if all such occurrences can be updated as well.

spuiuk: Yes. but not in this set of changes.

PR #74, opened 2024-03-25
Use python module pysmbc instead of smbclient

spuiuk:
This change introduces a new class SMBClient which is a wrapper around the python module pysmbc. This simplifies access to samba server as demonstrated in the changes to the consistency test which used to use smbclient to access the server.

Note that this changeset consists of only one commit and the rest of the commits are from #73

depends on #73
depends on https://github.com/samba-in-kubernetes/sit-environment/pull/93

### DIFF
--- a/testcases/consistency/test_consistency.py
+++ b/testcases/consistency/test_consistency.py
@@ -7,72 +7,52 @@
 # ip addresses).
 
 import testhelper
+from testhelper import SMBClient
 import os
 import pytest
 import typing
-from pathlib import Path
-
 
 test_string = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
 test_info_file = os.getenv("TEST_INFO_FILE")
 test_info = testhelper.read_yaml(test_info_file)
 
 
-def file_content_check(f: typing.IO, comp_str: str) -> bool:
-    read_data = f.read()
-    return read_data == comp_str
-
-
-def consistency_check(mount_point: Path, ipaddr: str, share_name: str) -> None:
+def consistency_check(hostname: str, share_name: str) -> None:
     mount_params = testhelper.get_mount_parameters(test_info, share_name)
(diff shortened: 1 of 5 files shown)


testcases/consistency/test_consistency.py
```diff
@@ -7,72 +7,54 @@
...
+    # file read cycle
+    scon = SMBClient(
+        hostname,
+        share_name,
+        mount_params["username"],
+        mount_params["password"],
+    )
+    retstr = scon.simple_read(test_filename)
+    scon.unlink(test_filename)
+    del scon
```
phlogistonjohn, line 43: Using `del` on objects directly is a bit of a (small) code smell in my book. If the point here is to free up resources (the smbc connection?) then it would be better to make SMBClient (also) be a context manager and write the code like:
```python
with SMBClient(host, share, user, pass) as scon:
    scon.simple_write(abc, xyz)
```

If I right about the reason for the `del` and you want tips on making this a context manager, let me know. otherwise: https://realpython.com/python-with-statement/

spuiuk: yes. I put the del in to force the connection to be disconnected after each read or write cycle.

spuiuk: Added a disconnect() method to the SMBClient class. This eventually does a del on the smb context. 

We still cannot test this on the sit-environment because we can't install pysmbc on the client machine. But I am considering using another library which is selfcontained and entirely written in python.

spuiuk:
PR #76 opened which uses a different python smb client module which should be easier to install and use on our test systems. Leaving both open as we decide which module to go ahead with.

spuiuk:
Closing this PR in favour of PR #76 

The pysmbc module is the python bindings for the libsmb client library. This means that when installing using pip, we need to recompile the module. This requires additional development packages installed on the test systems. Instead we use the pysmb module which is a pure python module which is easier to use with the test system.

PR #76, opened 2024-03-28
testhelper: introduce class SMBClient - pysmb version

spuiuk:
This helper class SMBClient provides us with a wrapper around the pysmb module to help replace the smbclient binary.

Unlike PR #74, we use the pysmb module here instead which is a pure python module and is therefore easier to install on our test systems.

### DIFF
--- a/testcases/consistency/test_consistency.py
+++ b/testcases/consistency/test_consistency.py
@@ -7,72 +7,52 @@
 # ip addresses).
 
 import testhelper
+from testhelper import SMBClient
 import os
 import pytest
 import typing
-from pathlib import Path
-
 
 test_string = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
 test_info_file = os.getenv("TEST_INFO_FILE")
 test_info = testhelper.read_yaml(test_info_file)
 
 
-def file_content_check(f: typing.IO, comp_str: str) -> bool:
-    read_data = f.read()
-    return read_data == comp_str
-
-
-def consistency_check(mount_point: Path, ipaddr: str, share_name: str) -> None:
+def consistency_check(hostname: str, share_name: str) -> None:
     mount_params = testhelper.get_mount_parameters(test_info, share_name)
-    mount_params["host"] = ipaddr
-    try:
-        test_file = testhelper.get_tmp_file(mount_point)
-        test_file_resp = test_file.with_suffix(".resp")
-        test_file_remote = "test-" + ipaddr + "." + share_name
-        with open(test_file, "w") as f:
-            f.write(test_string)
-        put_cmds = "put  %s %s" % (test_file, test_file_remote)
-        (ret, output) = testhelper.smbclient(mount_params, put_cmds)
-        assert ret == 0, "Failed to copy file to server"
+    test_filename = "/test_consistency"
 
-        # The file read cycle
-        get_cmds = "get %s %s; rm %s" % (
-            test_file_remote,
-            test_file_resp,
-            test_file_remote,
-        )
-        (ret, output) = testhelper.smbclient(mount_params, get_cmds)
-        assert ret == 0, "Failed to copy file from server"
-        with open(test_file_resp, "r") as f:
-            assert file_content_check(
-                f, test_string
-            ), "File content does not match"
-    finally:
-        if test_file.exists():
-            test_file.unlink()
-        if test_file_resp.exists():
-            test_file_resp.unlink()
+    # file write cycle
+    smbclient = SMBClient(
+        mount_params["host"],
+        mount_params["share"],
+        mount_params["username"],
+        mount_params["password"],
+    )
+    smbclient.write_text(test_filename, test_string)
+    smbclient.disconnect()
+
+    # file read cycle
+    smbclient = SMBClient(
+        mount_params["host"],
+        mount_params["share"],
+        mount_params["username"],
+        mount_params["password"],
+    )
+    retstr = smbclient.read_text(test_filename)
+    smbclient.unlink(test_filename)
+    smbclient.disconnect()
+
+    assert retstr == test_string, "File content does not match"
 
 
-def generate_consistency_check(
-    test_info_file: dict,
-) -> typing.List[typing.Tuple[str, str]]:
-    if not test_info_file:
-        return []
+def generate_consistency_check() -> typing.List[typing.Tuple[str, str]]:
     arr = []
-    for share in testhelper.get_exported_shares(test_info):
-        s = testhelper.get_share(test_info, share)
-        arr.append((s["server"], s["name"]))
+    for sharename in testhelper.get_exported_shares(test_info):
+        share = testhelper.get_share(test_info, sharename)
+        arr.append((share["server"], share["name"]))
     return arr
 
 
-@pytest.mark.parametrize(
-    "ipaddr,share_name", generate_consistency_check(test_info)
-)
-def test_consistency(ipaddr: str, share_name: str) -> None:
-    tmp_root = testhelper.get_tmp_root()
-    mount_point = testhelper.get_tmp_mount_point(tmp_root)
-    consistency_check(mount_point, ipaddr, share_name)
-    os.rmdir(mount_point)
-    os.rmdir(tmp_root)
+@pytest.mark.parametrize("hostname,share_name", generate_consistency_check())
+def test_consistency(hostname: str, share_name: str) -> None:
+    consistency_check(hostname, share_name)
--- a/testhelper/smbclient.py
+++ b/testhelper/smbclient.py
@@ -1,0 +1,77 @@
+from smb.SMBConnection import SMBConnection  # type: ignore
+from smb import smb_structs, base  # type: ignore
+import typing
+import io
+
+
+class SMBClient:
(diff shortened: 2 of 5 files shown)


testcases/consistency/test_consistency.py
```diff
@@ -7,72 +7,52 @@
...
+
+    # file write cycle
+    scon = SMBClient(
+        hostname,
+        share_name,
+        mount_params["username"],
+        mount_params["password"],
+    )
+    scon.simple_write(test_filename, test_string)
+    scon.disconnect()
```
phlogistonjohn, line 32: I still think it would be a lot nicer to have your new class be capable as acting as a context manager. Then you can write:
```
with scon:
    scon.simple_write(test_filename, test_sring)
```
for example.

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+            self.error = True
+            raise
+
+    def rmdir(self, dpath: str) -> None:
+        self.ctx.deleteDirectory(self.share, dpath)
+
+    def unlink(self, fpath: str) -> None:
+        self.ctx.deleteFiles(self.share, fpath)
+
+    def simple_write(self, fpath: str, teststr: str) -> None:
```
phlogistonjohn, line 61: Naming suggestion: write_text similar to pathlib.Path function of similar operation.

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+        file_obj = io.BytesIO(teststr.encode())
+        try:
+            self.ctx.storeFile(self.share, fpath, file_obj)
+        except smb.smb_structs.OperationFailure:
+            self.error = True
+            raise
+        finally:
+            file_obj.close()
+
+    def simple_read(self, fpath: str) -> str:
```
phlogistonjohn, line 71: Naming suggestion: read_text similar to pathlib.Path function of similar operation.

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+
+class SMBClient:
+    """Use pysmb to access the SMB server"""
+
+    def __init__(self, hostname: str, share: str, username: str, passwd: str):
+        self.server = hostname
+        self.share = share
+        self.username = username
+        self.password = passwd
+        self.error = False
```
phlogistonjohn, line 15: What is `self.error` supposed to do that raising exceptions from the methods wouldn't accomplish?

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+                self.username,
+                self.password,
+                "smbclient",
+                self.server,
+                use_ntlm_v2=True,
+            )
+            self.ctx.connect(self.server)
+            self.connected = True
+        except smb.base.SMBTimeout:
+            self.error = False
```
phlogistonjohn, line 31: See above question regarding `self.error`. Why hide the exc if you are not going to internally do retries?

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def simple_read(self, fpath: str) -> str:
+        readobj = io.BytesIO()
+        try:
+            self.ctx.retrieveFile(self.share, fpath, readobj)
+        except smb.smb_structs.OperationFailure:
+            self.error = True
+            raise
+        else:
+            ret = readobj.getvalue().decode("ascii")
```
phlogistonjohn, line 79: Ah! Please don't use ascii encoding. Use utf8!

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+            file_obj.close()
+
+    def simple_read(self, fpath: str) -> str:
+        readobj = io.BytesIO()
+        try:
+            self.ctx.retrieveFile(self.share, fpath, readobj)
+        except smb.smb_structs.OperationFailure:
+            self.error = True
+            raise
+        else:
```
phlogistonjohn, line 78: `else` in a try-finally is a bit of a code smell. What is meant to do?

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+            ret = readobj.getvalue().decode("ascii")
+        finally:
+            readobj.close()
+        return ret
```
phlogistonjohn, line 82: Here's a suggestion to simplify the whole func:  
```python
with io.BytesIO() as readobj:
   self.ctx.retrieveFile(self.share, fpath, readobj)
   txt = readobj.getvalue().decode("utf8")
return txt
```

If you absolutely need to catch the OperationFailure too:
```python
try:
    with io.BytesIO() as readobj:
        self.ctx.retrieveFile(self.share, fpath, readobj)
        txt = readobj.getvalue().decode("utf8")
except smb.smb_structs.OperationFailure:
    # do stuff with the exception....
    self.error = True
    raise
return txt
```

testhelper/smbclient.py
phlogistonjohn: My comments about simplifying the read function also largely apply here too.

testcases/consistency/test_consistency.py
```diff
@@ -7,72 +7,52 @@
...
-
-def generate_consistency_check(
-    test_info_file: dict,
-) -> typing.List[typing.Tuple[str, str]]:
-    if not test_info_file:
-        return []
+    test_filename = "/test_consistency"
+
+    # file write cycle
+    scon = SMBClient(
```
synarete, line 25: Better rename to `smb_clnt` (I was almost confused with [scons](https://scons.org/) build system)

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
+from smb.SMBConnection import SMBConnection  # type: ignore
```
synarete, line 1: Why this `from import` when you `import smb` on next line?

testhelper/smbclient.py
synarete, line 31: This is confusing; I suspect most Python reader would want some exception to be raise instead of any other error-handling mechanism.

testcases/consistency/test_consistency.py
spuiuk: I thought about this and in my opinion it is nice to have but not really needed. We can revisit this later.

testhelper/smbclient.py
spuiuk: done

testhelper/smbclient.py
spuiuk: done

testhelper/smbclient.py
spuiuk: SMBClient.error now removed.

testhelper/smbclient.py
spuiuk: fixed.

testhelper/smbclient.py
spuiuk: fixed.

testhelper/smbclient.py
spuiuk: fixed. Agree - This is a much nicer way.

testcases/consistency/test_consistency.py
spuiuk: Changed.

testhelper/smbclient.py
spuiuk: Both switch to from import.

testhelper/smbclient.py
spuiuk: yeah. The SMBClient.error was used for reporting purposes on the loading test. However I agree, it is better to raise these exceptions and let the consumer handle it. One of the issue was I didn't want to raise smb.* exceptions which means the caller would then need to import the smb module. I now catch these module specific exceptions and raise a generic IOError instead.

testhelper/smbclient.py
phlogistonjohn: OK, that approach seems reasonable. Other approaches I could suggest are:
Import the OperationFailure error into your module either directly or subclass it.
Subclass IOError with a more specific type for your new module.
For now, what you are suggesting is fine.

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+                self.password,
+                "smbclient",
+                self.server,
+                use_ntlm_v2=True,
+            )
+            self.ctx.connect(self.server)
+            self.connected = True
+        except base.SMBTimeout as error:
+            print(error)
+            raise IOError("Error: connect")
```
phlogistonjohn, line 33: I won't block you on this but I don't like seeing print()s in "library" code.
```python
raise IOError(f"failed to connect: {error}")
```
instead perhaps?

testhelper/smbclient.py
```diff
@@ -0,0 +1,82 @@
...
+    def disconnect(self) -> None:
+        self.connected = False
+        self.ctx.close()
+
+    def readdir(self, path: str = "/") -> typing.List:
+        try:
+            dentries = self.ctx.listPath(self.share, path)
+        except smb_structs.OperationFailure as error:
+            print(error)
+            raise IOError("Error: readdir")
```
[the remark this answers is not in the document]
spuiuk: fixed

testhelper/smbclient.py
spuiuk: Changed to this.

phlogistonjohn (approved):
looka ok to me

synarete (approved):
lgtm

PR #79, opened 2024-05-01
Supplementary tests

spuiuk:
Add a supplementary group test. 

 This test is to check writes to a folder owned by a supplementary group.

Requirements:
1. username in share["user"] exists
2. username is part of supplementary group provided in
    user["extra"]["supplementary_group"]
3. share["backend"]["path"] should be set

Steps:
1. Create folder test_subdir/ on direct path with
   group set to sgroup and mode 0770
2. Upload file to test_subdir/test-cp

To allow us to run this test, we need to 
a) Allow direct access to the underlying filesystem - Through the share->backend->path conf config option.
b) Allow us to pass an extra configuration option through the extra->supplementary_group conf option.


anoopcs9:
Isn't #63 also related?

spuiuk:
> Isn't #63 also related?

yes, This PR replaces #63. I was waiting for the tests to complete before updating that PR.  I will be closing PR #63 in favour of this one.

There has been several changes in the rest of the code which means a big rewrite for supplementary tests were needed. This PR is a result of these changes.

PR #82, opened 2024-06-06
smbtorture: Add Shadow Copy related tests

Shwetha-Acharya:


testcases/smbtorture/selftest/flapping.glusterfs
```diff
@@ -10,3 +10,6 @@
 
 #https://github.com/gluster/samba-integration/issues/241
 ^samba3.smb2.deny.*
+
+# Strictly checking ioctl tests against cephFS only
```
obnoxxx, line 14: How does dropping a comment in the flapping.glusterfs file restrict these test to run against cephfs?

testcases/smbtorture/selftest/flapping.gpfs
```diff
@@ -5,3 +5,6 @@
 # Ignore due to lack of proper multichannel setup.
 ^samba3.smb2.session.bind2
 ^samba3.smb2.session.two_logoff
+
+# Strictly checking ioctl tests against cephFS only
```
obnoxxx, line 9: same question as before: how does a comment to `flapping.gpfs` restrict tests to cephfs?

obnoxxx (commented):
Thanks for the PR, @Shwetha-Acharya !

I have a few questions inline in the patch. Additionally, on a higher  level, I am not quite clear if the patch actually does what the title states: The title says that the patch adds shadow-copy related tests, but the patch seems to deal mostly with ioctl and sparse-file tests.

Please explain!

Shwetha-Acharya:
> Thanks for the PR, @Shwetha-Acharya !
> 
> I have a few questions inline in the patch. Additionally, on a higher level, I am not quite clear if the patch actually does what the title states: The title says that the patch adds shadow-copy related tests, but the patch seems to deal mostly with ioctl and sparse-file tests.
> 
> Please explain!

Hi @obnoxxx Thanks for the review. Basically ioctl.c contains many ioctl and sparse file related tests as you have rightly pointed out but it also includes tests like https://github.com/samba-team/samba/blob/master/source4/torture/smb2/ioctl.c#L42

I thought it would be better to include all tests in ioctl.c  in our test suite as they might be utilized while accessing/querying the snapshots. And sparse files can be snapshotted as well.

Their grouping together in smbtorture inside ioctl.c also lead me to similar speculations.

However, I am learning about the sparse files and ioctl through this PR itself. If you have any suggestions regarding the same, please let me know.

testcases/smbtorture/selftest/flapping.gpfs
Shwetha-Acharya: > same question as before: how does a comment to `flapping.gpfs` restrict tests to cephfs?

Same answer as above

obnoxxx:
> > Thanks for the PR, @Shwetha-Acharya !
> > I have a few questions inline in the patch. Additionally, on a higher level, I am not quite clear if the patch actually does what the title states: The title says that the patch adds shadow-copy related tests, but the patch seems to deal mostly with ioctl and sparse-file tests.
> > Please explain!
> 
> Hi @obnoxxx Thanks for the review. Basically ioctl.c contains many ioctl and sparse file related tests as you have rightly pointed out but it also includes tests like https://github.com/samba-team/samba/blob/master/source4/torture/smb2/ioctl.c#L42
> 
> I thought it would be better to include all tests in ioctl.c in our test suite as they might be utilized while accessing/querying the snapshots. And sparse files can be snapshotted as well.

Right, but  this is more of a coincidence and the real link between ioctl tests and shadow copies is that the shadow copy operations are implemented in SMB2 as ioctl calls. So the ioctl test suite also contains shadow copy tests.

> However, I am learning about the sparse files and ioctl through this PR itself.

That sounds like the perfect opportunity!

>  If you have any suggestions regarding the same, please let me know.

Meanwhile I reviewed the changes enough to be convinced that the PR will be ready to be moved out of draft status once the CI checks are passing more.

So please try and fix the CI failures is the only real request at this time.

obnoxxx (commented):
Another thought:

Instead of  filling just knownfail and flapping with tests, we could consider using the skip list that samba's selftest understands to really prevent tests from being executed at all.

Adding this mechanism would be a separate PR though that this one should be based on.

Shwetha-Acharya:
> Another thought:
> 
> Instead of filling just knownfail and flapping with tests, we could consider using the skip list that samba's selftest understands to really prevent tests from being execured at all.
> 
> Adding this mechanism would be a separate PR though that this one should be based on.

sure! that sounds like a good plan

anoopcs9:
> Another thought:
> 
> Instead of filling just knownfail and flapping with tests, we could consider using the skip list that samba's selftest understands to really prevent tests from being execured at all.

IIRC, we introduced dedicated flapping list for each backend in order to keep the _selftest_ subdirectory, which is largely a clone from upstream samba sources, intact. We occasionally rebase/sync the contents from upstream to keep us updated.

> Adding this mechanism would be a separate PR though that this one should be based on.

Having said that I am not sure why we didn't sync the _skip_ list from upstream.

@spuiuk Any further thoughts?

spuiuk:
We don't use the skip list because we don't use the test.py file which is called by "make tests" to generate a list of tests to be run.  We instead list out the testsuites we want to run.

To work around these problems, I would instead of adding the entire smb2.ioctl test suite, simply add tests which I would like to run.

spuiuk:
> Another thought:
> 
> Instead of filling just knownfail and flapping with tests, we could consider using the skip list that samba's selftest understands to really prevent tests from being executed at all.
> 
> Adding this mechanism would be a separate PR though that this one should be based on.

We do not have a skip list since we do not copy over the entire selftest directory and use only those parts which we need from the selftest directory. The skip list are used by make test to generate a list of tests to run. We took the easy approach which simply involved listing the tests we want to run.

This smbtorture test suite was built to run against the a single filesystem - glusterfs. When we added newer filesystems, we discovered that we had some tests which passed against one filesystem but failed against another. We needed to investigate these failures but wanted to see passing tests - so we add the failed tests to the issue tracker to investigate later and add the failing tests to the filesystem specific flapping list.

We have never encountered a uxsuccess until now. So the easiest workaround is to add the remaining tests we want to check for but avoid listing the tests which do a uxsuccess for now.

Shwetha-Acharya:
> > Another thought:
> > Instead of filling just knownfail and flapping with tests, we could consider using the skip list that samba's selftest understands to really prevent tests from being executed at all.
> > Adding this mechanism would be a separate PR though that this one should be based on.
> 
> We do not have a skip list since we do not copy over the entire selftest directory and use only those parts which we need from the selftest directory. The skip list are used by make test to generate a list of tests to run. We took the easy approach which simply involved listing the tests we want to run.
> 
> This smbtorture test suite was built to run against the a single filesystem - glusterfs. When we added newer filesystems, we discovered that we had some tests which passed against one filesystem but failed against another. We needed to investigate these failures but wanted to see passing tests - so we add the failed tests to the issue tracker to investigate later and add the failing tests to the filesystem specific flapping list.
> 
> We have never encountered a uxsuccess until now. So the easiest workaround is to add the remaining tests we want to check for but avoid listing the tests which do a uxsuccess for now.

That is fair. Tracking all these failures becomes a task considering existence of various file systems and different combinations of failures.

spuiuk:
> That is fair. But, tracking all these failures becomes a task considering existence of various file systems and different combinations of failures when we only add the tests we want to run.

The adding of the failing tests to the flapping lists is expected to be temporary while we investigate the failures and fix them.

spuiuk:
uxsuccess failures are reported when the test passes but is listed in the knownfail and expected failure lists. 

```
knownfail.d/sparse_files:^samba3.smb2.ioctl.sparse_punch
knownfail.d/sparse_files:^samba3.smb2.ioctl.sparse_qar_ob1
knownfail.d/sparse_files:^samba3.smb2.ioctl.copy-chunk streams
```

causes the following failure reported.
```
uxsuccess: samba3.smb2.ioctl.sparse_punch_invalid
```

anoopcs9:
> ## cephfs:
> 
> Here also there is one unrelated failure related smb2.charset.

We already have an issue(https://github.com/samba-in-kubernetes/sit-test-cases/issues/91) reported.

> Respective ioctl failures of kclient are already added as flapping under cephfs However many other ioctl copy-chunk related tests are failing for cephfs-vfs and cephfs-vfs.new
> 
> test_smbtorture[share-cephfs-default-vfs-smb2.charset] – FAILED
> test_smbtorture[share-cephfs-default-kclient-smb2.ioctl] – FAILED

This is a failure on kclient.

> test_smbtorture[share-cephfs-default-vfs-smb2.ioctl] – FAILED
> test_smbtorture[share-cephfs-default-vfs.new-smb2.ioctl] – FAILED

As we can see _smb2.ioctl_ test suite still has some issues with CephFS(in any mode).

Shwetha-Acharya:
> > ## cephfs:
> > Here also there is one unrelated failure related smb2.charset.
> 
> We already have an issue(#91) reported.
> 
> > Respective ioctl failures of kclient are already added as flapping under cephfs However many other ioctl copy-chunk related tests are failing for cephfs-vfs and cephfs-vfs.new
> > test_smbtorture[share-cephfs-default-vfs-smb2.charset] – FAILED
> > test_smbtorture[share-cephfs-default-kclient-smb2.ioctl] – FAILED
> 
> This is a failure on kclient.
> 
> > test_smbtorture[share-cephfs-default-vfs-smb2.ioctl] – FAILED
> > test_smbtorture[share-cephfs-default-vfs.new-smb2.ioctl] – FAILED
> 
> As we can see _smb2.ioctl_ test suite still has some issues with CephFS(in any mode).

yes correct, with kclient, sparse file tests are failing, and copy chunk stream test is failing.
While on vfs and vfs.new almost 15 copy chunk related tests are failing.

PR #84, opened 2024-06-18
testhelper/smbclient.py: use smbprotocol python module

spuiuk:
pysmb only support SMB2.0 version. The samba-container defaults to using SMB2.1 protocol and we are unable to use pysmb to connect to the samba server. This is a major issue.

The smbprotocol python module avoids this problem.

### DIFF
--- a/testhelper/smbclient.py
+++ b/testhelper/smbclient.py
@@ -1,111 +1,137 @@
-from smb.SMBConnection import SMBConnection  # type: ignore
-from smb import smb_structs, base  # type: ignore
+from smbprotocol.exceptions import SMBException  # type: ignore
+import smbclient  # type: ignore
 import typing
-import io
+
+rw_chunk_size = 1 << 21  # 2MB
 
 
 class SMBClient:
-    """Use pysmb to access the SMB server"""
+    """Use smbprotocol python module to access the SMB server"""
 
-    def __init__(self, hostname: str, share: str, username: str, passwd: str):
+    def __init__(
+        self,
+        hostname: str,
+        share: str,
+        username: str,
+        passwd: str,
+        port: int = 445,
+    ):
         self.server = hostname
         self.share = share
-        self.username = username
-        self.password = passwd
+        self.port = port
+        self.client_params = {
+            "username": username,
+            "password": passwd,
+            "connection_cache": {},
+        }
+        self.prepath = f"\\\\{self.server}\\{self.share}\\"
         self.connected = False
         self.connect()
+
+    def _path(self, path: str = "/") -> str:
+        path.replace("/", "\\")
+        return self.prepath + path
 
     def connect(self) -> None:
         if self.connected:
             return
         try:
-            self.ctx = SMBConnection(
-                self.username,
-                self.password,
-                "smbclient",
-                self.server,
-                use_ntlm_v2=True,
+            smbclient.register_session(
+                self.server, port=self.port, **self.client_params
             )
-            self.ctx.connect(self.server)
             self.connected = True
-        except base.SMBTimeout as error:
+        except SMBException as error:
             raise IOError(f"failed to connect: {error}")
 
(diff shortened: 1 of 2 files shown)


testhelper/smbclient.py
```diff
@@ -1,111 +1,134 @@
...
+        username: str,
+        passwd: str,
+        port: int = 445,
+    ):
         self.server = hostname
         self.share = share
-        self.username = username
-        self.password = passwd
+        self.port = port
+        self.client = {
```
phlogistonjohn, line 23: Minor nitpick: I was skimming though the code and saw something like `smbclient.open_file(path, mode=xyz, **self.client)` and was confused and went back to see that a client was indeed a dict. I suggest renaming this var something like `client_params` or `client_options` for clarity.

spuiuk: Thanks. Fixed it.

testhelper/smbclient.py
```diff
@@ -1,111 +1,138 @@
...
-            raise IOError(f"failed in write_text: {error}")
-        except base.SMBTimeout as error:
-            raise TimeoutError(f"write_text: {error}")
-        except base.NotConnectedError as error:
-            raise ConnectionError(f"write: {error}")
+            with smbclient.open_file(
+                self._path(fpath), mode="wb", **self.client_params
+            ) as fd:
+                while True:
+                    data = writeobj.read(4096)
```
synarete, line 103: Why suck small chunk size? This is inefficient. Modern machines should have no problem to do it with 1M, 2M or 4M chunk sizes.

testhelper/smbclient.py
```diff
@@ -1,111 +1,138 @@
...
+                self._path(fpath), mode="rb", **self.client_params
+            ) as fd:
+                while True:
+                    data = fd.read(4096)
+                    if len(data) == 0:
+                        break
+                    n = 0
+                    while n < len(data):
+                        n += readobj.write(data[n:])
+        except SMBException as error:
```
synarete, line 125: This whole read-write in chunks logic should be separated into helper function. Same hold for lines 93-99 above. In addition, the `if len(data) == 0` may be removed as it is already handled implicitly by the following `while` loop.

spuiuk: I have split out the read-write cycle into a helper function.
The if condition is needed to break out of the while loop which reads data before writing to the fd.

I have uploaded the new set of code.  Can you please take a look. Thanks.

testhelper/smbclient.py
```diff
@@ -1,111 +1,137 @@
-from smb.SMBConnection import SMBConnection  # type: ignore
-from smb import smb_structs, base  # type: ignore
+from smbprotocol.exceptions import SMBException  # type: ignore
+import smbclient  # type: ignore
 import typing
-import io
+
+rw_chunk_size = 1 << 21  # 21MB
```
synarete, line 7: Nit: 2MB (not 21)

synarete (approved):
Looks better. See nit comment.

testhelper/smbclient.py
spuiuk: fixed.

anoopcs9 (changes requested):

Shwetha-Acharya (commented):
except for the line 11 change, commented about already, rest of the things look good to me

testhelper/smbclient.py
```diff
@@ -1,111 +1,146 @@
...
+from smbprotocol.connection import Connection  # type: ignore
 import typing
-import io
+import uuid
+
+rw_chunk_size = 1 << 21  # 2MB
 
 
 class SMBClient:
     """Use pysmb to access the SMB server"""
```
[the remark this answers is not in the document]
spuiuk: Thanks - Good catch. Have fixed it in the latest submission.

anoopcs9 (approved):

PR #87, opened 2024-07-05
Add smb2.session.reauth4 to flapping list for XFS and CephFS(kclient)

anoopcs9:
With recent changes to _[smb.conf](https://github.com/samba-in-kubernetes/sit-environment/pull/109)_, _smb2.session.reauth4_ is expected to fail on XFS and CephFS kclient method. You can find more details from https://github.com/samba-in-kubernetes/sit-environment/pull/109#issuecomment-2208737127.

Also remember that this is **_not a problem with CephFS vfs_** approach but our clone of selftest mechanism currently lacks the ability to consider flapping lists based on variants for a particular backend.

### DIFF
--- a/testcases/smbtorture/selftest/flapping.cephfs
+++ b/testcases/smbtorture/selftest/flapping.cephfs
@@ -16,3 +16,7 @@ samba3.smb2.timestamps.time_t_10000000000
 samba3.smb2.timestamps.time_t_-1
 samba3.smb2.timestamps.time_t_-2
 samba3.smb2.timestamps.time_t_1968
+
+# https://github.com/samba-in-kubernetes/sit-environment/pull/109
+# Note: CephFS(vfs) successfully completes smb2.session.reauth4.
+samba3.smb2.session.reauth4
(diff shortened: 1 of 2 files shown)


spuiuk:
We could use "cephfs-vfs"(or any other variation) as a new filesystem.
The cost is simply a new flapping.cephfs-vfs list in the selftest directory which needs to be maintained. Since this doesn't change all that much, it isn't all that difficult to maintain.  I expect to see more of this in the future and therefore should be fixed properly now.

To do this.
1) Add a new patch to copy flapping.ceph list to a  new flapping.ceph-vfs list. This patch has to go in before we modify the file to add smb2.session.reauth4 to it.
Other patches from this PR apply.

We then
2) patch sit-environment to user "filesystem-variant" when setting the filesystem in testinfo.yml

anoopcs9:
>     1. Add a new patch to copy flapping.ceph list to a  new flapping.ceph-vfs list. This patch has to go in before we modify the file to add smb2.session.reauth4 to it.
>        Other patches from this PR apply.

#88 created.

spuiuk (approved):
ACK

PR #88, opened 2024-07-05
selftest: Add a dedicated flapping list for CephFS vfs

anoopcs9:
With certain samba configurations we might require a separate flapping list for variants of a backend. One such situation is detailed in https://github.com/samba-in-kubernetes/sit-environment/pull/109 where specific test run results differ among variants of a particular backend.


spuiuk (approved):
ACK

PR #89, opened 2024-07-08
test-info.yml: add filesystem variant to the backend

spuiuk:
This helps us tailor tests for cases where the behaviour changes according to the filesystem variant being used.

In this case, we use it with the smbtorture tests where we select the flapping list for the filesystem-variant and it it doesn't exist, fall back to filesystem file.

### DIFF
--- a/testcases/smbtorture/test_smbtorture.py
+++ b/testcases/smbtorture/test_smbtorture.py
@@ -12,12 +12,24 @@ from pathlib import Path
 
 script_root = os.path.dirname(os.path.realpath(__file__))
 smbtorture_exec = "/bin/smbtorture"
-filter_subunit_exec = script_root + "/selftest/filter-subunit"
-format_subunit_exec = script_root + "/selftest/format-subunit"
+selftest_dir = script_root + "/selftest"
+filter_subunit_exec = selftest_dir + "/filter-subunit"
+format_subunit_exec = selftest_dir + "/format-subunit"
 smbtorture_tests_file = script_root + "/smbtorture-tests-info.yml"
 
 test_info_file = os.getenv("TEST_INFO_FILE")
 test_info = testhelper.read_yaml(test_info_file)
+
+
+def flapping_file(backend: str, variant: str = "") -> str:
+    file = "flapping." + backend
+    if variant and variant != "default":
+        file_variant = f"{file}-{variant}"
+        if os.path.exists(os.path.join(selftest_dir, file_variant)):
+            return file_variant
+    if os.path.exists(os.path.join(selftest_dir, file)):
+        return file
+    return ""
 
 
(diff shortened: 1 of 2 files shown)


testcases/smbtorture/test_smbtorture.py
```diff
@@ -12,14 +12,26 @@
...
 test_info_file = os.getenv("TEST_INFO_FILE")
 test_info = testhelper.read_yaml(test_info_file)
 
 
+def flapping_file(backend: str, variant: str = "") -> str:
+    file = "flapping." + backend
+    if variant:
+        file_variant = f"{file}-{variant}"
+        if os.path.exists(os.path.join(selftest_dir, file_variant)):
+            return file_variant
```
anoopcs9, line 29: With https://github.com/samba-in-kubernetes/sit-environment/pull/111 we will always have variant defined in _test-info.yml_(currently either `default` or `vfs`). Thus we won't find files like _flapping.xfs-default_, _flapping.cephfs-default_ etc on the system and we return "" causing some expected failures affect the overall test result.

spuiuk: Thanks..  I'll make an exception for default in the code.

We need to handle 3 cases.
1) With variant not set. - use flapping.cephfs
2) With variant set - use flapping.cephfs-vfs
3) with variant set to "default" - use flapping.cephfs

testcases/smbtorture/test_smbtorture.py
```diff
@@ -12,14 +12,26 @@
...
 test_info_file = os.getenv("TEST_INFO_FILE")
 test_info = testhelper.read_yaml(test_info_file)
 
 
+def flapping_file(backend: str, variant: str = "") -> str:
+    file = "flapping." + backend
+    if variant and variant != "default":
+        file_variant = f"{file}-{variant}"
+        if os.path.exists(os.path.join(selftest_dir, file_variant)):
+            return file_variant
```
anoopcs9, line 29: Should we improvise ourselves by returning the default flapping list for a particular backend if the corresponding variant file is not present 😜 ?

anoopcs9 (approved):
lgtm.

testcases/smbtorture/test_smbtorture.py
spuiuk: actually. Good catch. I meant to fall back to using the non variant flapping list in case the variant file isn't there. the elif prevents that. It should be a straight forward if condition.

spuiuk: pushed in a new version of the code.

anoopcs9: Thanks.

anoopcs9 (approved):

PR #97, opened 2025-01-30
testcases/smbtorture: update selftest files

spuiuk:
Head of samba repo when this copy was made
062dc07e9b9 (s3-libnet: avoid using lp_dns_hostname() in join code, 2025-01-14)


testcases/smbtorture/selftest/knownfail.d/smb2.replay.glusterfs
anoopcs9, line 1: Was this intentional to remove _smb2.replay.glusterfs_? If not can we avoid removing it for the name sake?

anoopcs9 (changes requested):

testcases/smbtorture/selftest/knownfail.d/smb2.replay.glusterfs
spuiuk: yep. unintentional. Good catch. Thanks.

anoopcs9 (approved):
lgtm, thanks.

PR #107, opened 2025-09-22
smbtorture: Add `modern_write_time_update-1` test to flapping.cephfs

anoopcs9:
_smb2.timestamps.modern_write_time_update-1_ does not run successfully on non-proxy enabled CephFS shares due to an underlying bug which is reported via tracker [#71622](https://tracker.ceph.com/issues/71622). For now we add it to the flapping list until the issue is fixed.


spuiuk (approved):
ACK

PR #110, opened 2025-11-03
containers/ltp: Update the custom list of ltp tests

anoopcs9:
Recently ltp container image was rebuilt based on the latest sources. Accordingly update the list of tests as per the expectation based on the list of passing tests (from QE) on EL9 kernel versions. Few tests had to be excluded as their corresponding updated sources requires support for mknod, mkfifo, symlink etc. which are not available with the way we perform the kernel CIFS mount.


anoopcs9:
> Errors occurred during LTP tests:
/opt/ltp/testcases/bin/rename14

The [fix](https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/merge_requests/7537) for the above failure is still pending merge in CentOS Stream 9 kernel.

synarete (approved):
lgtm

spuiuk:
Overriding the failures and merging.

